Issue #8, opened 2019-10-11
Defaults read from config file


fatso83:
We have a `develop` branch as our default, which doesn't work that well with your default. Having a way of having that set as the branch default would remove a bit of manual git work that we have to do today.

A checked in .mobdefault would help :-)

simonharrer:
Hm, I am somewhat reluctant to add support for such files. Environment variables are available to configure the tool. Should work as long as you don't mob together with teams, each having their own "defaults". Not sure this is a typical use case, though.

simonharrer:
Closing it for now. Maybe will reopen it later, if more people have this kind of use case.

PR #9, opened 2019-10-11
Add pointers on how to configure `mob`

fatso83:
Slightly related to #8, with regards to how one can configure the process.


simonharrer:
Thanks for your contribution!

PR #11, opened 2019-10-30
Remove help description for `mob join`

L7R7:
The `join` option is gone, so it shouldn't appear in the help text


simonharrer:
Thanks. Forgot this one.

PR #16, opened 2020-03-13
Allow --no-verify when committing and pushing WIP commits

mcnally:
We have a few git hooks running when committing and pushing (lint/test etc), which caused this to fail when doing `mob next`.

This change allows for the -no-verify flag to be set on git commands during `mob next` so it skips the hooks and doesn't slow down or interrupt the handover.


simonharrer:
Thinking about it, it should probably be the other way round. Always us no-verify for mob next, but if you really want that, you could disable that. Why would one want to have pre-commit hooks anyway for a WIP commit?

I would get rid of the global variable wipGitFlags and compute through a function based on the flag variable the additonal parameter or not.

simonharrer:
Forgot to say the most important things: thanks for using 'mob' and for helping out making it even better! Please let me know what you think of my two comments above. Looking forward to your reply!

simonharrer:
We always use "--no-verify". No config option required. Closing this PR as already merged in #20 .

Issue #14, opened 2019-11-21
mob start should start from the current branch instead of master


alexkli:
# Current behavior
Currently, `mob start` will start the `mob-session` branch from `master`, regardless of what the currently checked out branch is. It is possible to control the base branch using `MOB_BASE_BRANCH=mybranch mob start`. However, this is non-intuitive and hard to remember.

# Suggestion

`mob start` starts from the current checked out branch (if no `mob-session` exists and no local modifications, etc. same checks as before). You can still specify `MOB_BASE_BRANCH` to override that if you want.

The output would make it clear what branch was used and also tell the user how to undo, in case they did not want to start from the current branch, i.e. using `mob done` or `mob reset`.

# Reasoning

Is there a particular reason or use case why it ignores the current branch and always starts `mob-session` based off of master?

In our experience, when you start the mob, you want to start it from your current branch. You might have already looked at the code before you actually run the `mob start` command, so there is IMO little chance that you want to start it from another branch (`master`). It also feels more natural to do

```
git checkout myfeature
mob start
```

than 

```
MOB_BASE_BRANCH= myfeature mob start
```

Users will know git a lot better than the mob tool.

To given an example, we have been working on a feature branch that represents a more complex effort which we handle in the mob. This took many days and we don't necessarily do a mob every day. That means we split up some work in the meantime, contributed that back to the feature branch and the next day would do a mob and work on the integration. All the time the work focuses on that feature branch, not master.

OTOH, I feel that `mob-session` is not a feature branch, but just a "the mob is working on this today" temporary working branch, as there can only be 1, you don't really want to do a PR from `mob-session` and it's easy to loose the work using `mob done` or even `mob reset`.

simonharrer:
Thanks for the issue. 

> Is there a particular reason or use case why it ignores the current branch and always starts mob-session based off of master?

We, i.e. my team, only works in master. We don't use any other branch, except on exceptions. We found this is the best way when working full-time remote mob programming. I want to support this way.

 I see your point in doing mob programming only for specific feature branches as you don't do mob programming full-time. Will have to think about the consequences in changing the current behaviour though. To work correctly, the information of the base branch would have to be stored somehow, making the handling more complex, because how does mob done know into which branch it should revert back?

> I feel that mob-session is not a feature branch, but just a "the mob is working on this today" temporary working branch

That's how we think of it. May I add your description of the 'mob-session' branch to the README?

alexkli:
[...]

PR #17, opened 2020-03-14
Use the currently checked-out branch as the baseBranch

MStausberg:
instead of using the default "master" or the environment variable MOB_BASE_BRANCH, we get the currently checked-out branch from git and set it as the baseBranch.

Solves #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 Swift handover for remote mobs using git.
 `mob` is a CLI tool written in GO.
-It keeps your master branch clean and creates WIP commits on `mob-session` branch.
+It keeps your working branch clean and creates WIP commits on `mob-session` branch.
 
 ## How to use it?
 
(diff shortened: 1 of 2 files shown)


koppor:
@MStausberg Could you state why this PR was closed? I think, it's a good solution. Maybe the user prompt could be added:

> The branch "feature-123" is used as mob-base branch. Do you want to continue?

MStausberg:
Yeah I realized I made a logical error, the solution I implemented would set the base branch to the current branch whenever any mob-command is executed.
So it would work fine when starting, but afterwards it would change the base branch to the mob-branch.

simonharrer:
I'm working on v1.0.0 right now in https://github.com/remotemobprogramming/mob/pull/38 and it may help in providing a solution to a more flexible "start" branch and a "done" branch.

PR #18, opened 2020-03-16
Run the install script in the correct environment

martinotten:
My default shell is not bash compatible and the install script failed. The shebang would guarantee a correct interpreter being used.


simonharrer:
Awesome. Thanks for your contribution!

PR #19, opened 2020-03-19
Configurable voice output program

froth:
My operating system does not provide say but an alternative called espeak. I therefore made the command to use for text output configurable.


simonharrer:
❤️ great PR thanks for contrib!

PR #20, opened 2020-03-23
Always commit the "mob next" commit with --no-verify to skip pre-commit hooks

kyusu:
Hi @simonharrer,

thanks for creating the remote mob programming tool.

When we started using it, we quickly ran into the issue that the `mob next` commits were rejected by our rather strict git pre-commit hook. Since we don't want to get rid of the hook and you mentioned [here](https://github.com/remotemobprogramming/mob/pull/16#issuecomment-599636689) that you agree that always using `--no-verify` makes a lot of sense for work-in-progress commit, I've add the (hard-coded) flag.  Making it a configurable option is imho not worth the added complexity.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -190,7 +190,7 @@ func next() {
 		sayInfo("nothing was done, so nothing to commit")
 	} else {
 		git("add", "--all")
-		git("commit", "--message", "\""+wipCommitMessage+"\"")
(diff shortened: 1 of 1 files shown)


simonharrer:
Fine for me. If someone has issues with it, we can still add the config option. Thanks for helping out!

PR #22, opened 2020-04-01
Add linux support for share

mariusbuescher:
Actually there is a way to make the zoom screenshare work on Linux as well. It depends on `xdotool`.


simonharrer:
Awesome. Thanks for your contribution!

PR #23, opened 2020-04-02
Only print "success" when successfully installed

HarrisonMc555:
When I tried to install `mob`, I got a bunch of Go compiler warnings, but it still said it installed `mob`. This was pretty confusing.

This change conditions the "success" message on the successful compilation and copying of the compiled file.


simonharrer:
Thanks for your contribution. Your PR makes total sense. :-)

PR #29, opened 2020-04-22
Install in userspace

enolive:
rationale: I do not like to install volatile dev tools in the system space. So allow the script to install the tool also into the local bin folder if no sudo rights were given

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,15 @@ It keeps your master branch clean and creates WIP commits on `mob-session` branc
 ## How to install
 
 ```bash
-sh -c "$(curl -s https://raw.githubusercontent.com/remotemobprogramming/mob/master/install.sh)"
+curl -s https://raw.githubusercontent.com/remotemobprogramming/mob/master/install.sh | sh
 ```
+
+or, if you wish to install the tool for all users, add a `sudo`.
+
+```bash
+curl -s https://raw.githubusercontent.com/remotemobprogramming/mob/master/install.sh | sudo sh
+```
+
 
 ## How to use it?
 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Install that on Debian/Ubuntu/Mint as follows:
 sudo apt-get install espeak-ng-espeak mbrola-us1
 ```
 
-Create a little script in your $PATH with the following content:
+Create a little script in your $PATH called `say` with the following content:
 
 ```bash
 #!/bin/sh
--- a/README.md
+++ b/README.md
@@ -12,13 +12,6 @@ It keeps your master branch clean and creates WIP commits on `mob-session` branc
 curl -s https://raw.githubusercontent.com/remotemobprogramming/mob/master/install.sh | sh
 ```
 
-or, if you wish to install the tool for all users, add a `sudo`.
-
-```bash
-curl -s https://raw.githubusercontent.com/remotemobprogramming/mob/master/install.sh | sudo sh
-```
-
-
 ## How to use it?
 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -1,25 +1,98 @@
...
-
-location="$(which mob)"
-echo "Mob binary location: $location"
-
-version="$(mob version)"
-echo "Mob binary version: $version"
\ No newline at end of file
+#!/bin/sh   
+target=/usr/local/bin
+local_target=$(systemd-path user-binaries)
```
simonharrer, line 3: We should only execute this if the parameter to the script is "--user", otherwise, one might get strange error logs.

install.sh
```diff
@@ -1,25 +1,98 @@
...
+    if [ "$local_target" != "" ]
+    then
+      echo "we recommend that you use the --user flag"
+      echo "to install the app into your user binary path $local_target"
+      echo
+      echo "  $stream_cmd | sh -s - --user"
+      echo
+    fi
+    echo "calling the installation with sudo might help."
+    echo "please do it ONLY if you understand the implications of calling some remote script with ROOT rights."
```
simonharrer, line 40: Would remove this, people know that. With great power comes great responsibility. :-)

install.sh
```diff
@@ -1,25 +1,98 @@
...
+    echo
+    echo "  $stream_cmd | sudo sh"
+    echo
+    exit 1
+  fi
+}
+
+download_binary() {
+  echo "downloading latest 'mob' release from GitHub..."
```
simonharrer, line 49: It's more like installing latest 'mob' release from GitHub to $target

install.sh
```diff
@@ -1,25 +1,98 @@
...
+user_arg=$1
+stream_cmd="curl -s https://raw.githubusercontent.com/remotemobprogramming/mob/master/install.sh"
+readme_location="https://github.com/remotemobprogramming/mob/blob/master/README.md"
+
+handle_user_installation() {
+  if [ "$user_arg" = "--user" ]
+  then
+    if [ "$local_target" != "" ] && [ ! -d "$local_target" ]
+    then
+      mkdir $local_target
```
simonharrer, line 13: 'mkdir -p'?

simonharrer: Should the script fail if mkdir fails? Probably so.

install.sh
```diff
@@ -1,25 +1,98 @@
...
+check_say() {
+  say=$(which say)
+  if [ ! -e "$say" ]
+  then
+    echo
+    echo "you do not have an installed 'say' command on your system."
+    echo "while 'mob' will still work, you won't get any nice spoken indication that your time is up."
+    echo "please refer to the README.md how to setup text to speech on a *NIX system."
+    echo
+    echo "$readme_location#linux-timer"
```
simonharrer, line 85: Would put the '#linux-timer' to the readme location variable as well.

install.sh
enolive: if you look down, you'll see that the next command will fail as the directory won't exist.

README.md
```diff
@@ -36,10 +36,21 @@ The `mob share` feature only works if you activate make the screenshare hotkey i
...
-To get the timer to work on Linux, you need the GNUstep speech engine. Install that on ubuntu as follows:
+To get the timer to play "mob next" when your time is up, you'll need an installed speech engine. 
+Install that on Debian/Ubuntu/Mint as follows:
 
 ```bash
-sudo apt-get install gnustep-gui-runtime
+sudo apt-get install espeak-ng-espeak mbrola-us1
+```
+
+Create a little script in your $PATH with the following content:
```
simonharrer, line 46: Name of the script should be given as well. It should be named 'say'.

PR #32, opened 2020-05-07
Add security settings on macOS Catalina

nowsprinting:
Need security settings for use `osascript` on  macOS Catalina (or later, maybe).
However, it may be used in a malicious attack, so please decide whether to post it as it is.


simonharrer:
Haven't had to do anything like this. And it still worked fine. Can you describe a little bit more detailed why you had to enable that?

nowsprinting:
I got into the same situation as this.
https://stackoverflow.com/questions/54973241/applescript-application-is-not-allowed-to-send-keystrokes/58401185?stw=2#58401185

Maybe it only happens if update from older OS version to Catalina?

simonharrer:
Hm maybe. Might be a very rate case as I haven’t heard of such an issue since the start of this tool over a year ago. I would close this PR and reopen if more people have similar issues. Is that ok with you?

nowsprinting:
Of course!

PR #33, opened 2020-05-07
Windows git bash installation

enolive:
Rationale: 

While using the WSL is a nice alternative, not everyone can install it (company policies, availability of the hypervisor, etc). As most Windows devs should have a git installation (if not, they can't work with `mob` anyways :wink:) it would be nice support installing mob via the git-bash that comes bundled with it.

Pros:

* no extra effort required for maintaining a install.cmd script
* user can install mob very easily
* shell syntax much better than the horrible Batch syntax
* external tools in the installer such as tar, tr, curl also bundled within git-bash
* no GoLang binary required for installing mob

Cons: 

* user feels comfortable in Windows and does not move to a better OS, such as OSX or Linux :wink:  
* PowerShell might be a feasible alternative. But it has still the problem of maintaining 3 different installers (brew, *NIX sh, windows)

TODOs:

* [x] testing
* [x] consider if the install.cmd script should be removed entirely

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,6 @@ Swift [git handover](https://www.remotemobprogramming.org/#git-handover) with 'm
 - `mob` creates WIP commits on the `mob-session` branch
 - `mob` notifies you when it's time to handover
 
-[__TOC__]
-
 ## How to install
 
 ```
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ Swift [git handover](https://www.remotemobprogramming.org/#git-handover) with 'm
 - `mob` creates WIP commits on the `mob-session` branch
 - `mob` notifies you when it's time to handover
 
+[__TOC__]
+
 ## How to install
 
 ```
@@ -91,6 +93,15 @@ Create a little script in your `$PATH` called `say` with the following content:
 ```bash
 #!/bin/sh
 espeak -v us-mbrola-1 "$@"
+```
+
+eSpeak NG works even on a MINGW64 installation such as **Git-Bash under Windows**
+(assuming you have [installed the binary through their MSI](https://github.com/espeak-ng/espeak-ng/releases)):
+
+```bash
+#!/bin/sh
+# eSpeak does not set any path, so we specify its installation directory instead
+"$PROGRAMFILES/eSpeak NG/espeak-ng" "$@"
 ```
 
 ## How to configure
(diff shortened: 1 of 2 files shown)


simonharrer:
Works, but when the installation doesn't find a say command, it displays the info for linux and not for git bash on windows. We should add a specific info for that. Otherwise looks fine to me! :-)

simonharrer:
Not sure about the install.cmd. Maybe discuss this in another PR or issue?

README.md
```diff
@@ -93,6 +93,15 @@ Create a little script in your `$PATH` called `say` with the following content:
 espeak -v us-mbrola-1 "$@"
 ```
 
+eSpeak NG works even on a MINGW64 installation such as **Git-Bash under Windows**
```
simonharrer, line 96: Needs a section heading so we can reference to this from the install script

README.md
```diff
@@ -93,6 +93,15 @@ Create a little script in your `$PATH` called `say` with the following content:
...
+(assuming you have [installed the binary through their MSI](https://github.com/espeak-ng/espeak-ng/releases)):
+
+```bash
+#!/bin/sh
+# eSpeak does not set any path, so we specify its installation directory instead
+"$PROGRAMFILES/eSpeak NG/espeak-ng" "$@"
+```
```
simonharrer, line 103: Maybe tell the user to put this content in `~/bin/say` for it to work?

README.md
```diff
@@ -133,4 +142,4 @@ Contributions and testing by Jochen Christ, Martin Huber, Franziska Dessart, and
 Logo designed by [Sonja Scheungrab](https://twitter.com/multebaerr).
 
 <script async defer src="https://cdn.simpleanalytics.io/hello.js"></script>
-<noscript><img src="https://api.simpleanalytics.io/hello.gif" alt=""></noscript>
\ No newline at end of file
+<noscript><img src="https://api.simpleanalytics.io/hello.gif" alt=""></noscript>
```
simonharrer, line 145: Did you remove the trailing line break?

simonharrer:
I do like your way of writing a bash file. It's really very clear and understandable!

README.md
enolive: not intentionally. What is the problem with that?

simonharrer: Git works best with a trailing line break at the end of each file.

enolive:
> I do like your way of writing a bash file. It's really very clear and understandable!

thanks! I believe that deployment scripts are the most crucial part of the software (even more important than tests) and thus should have a high quality.

PR #43, opened 2020-06-08
implemented status command; also added a couple tests for its output

jlangr:
hi Simon--hope this is ok. Status command was basically just showing help. I split out an "execute" function in order to make it easy to test that a command was kicked off. I also added a couple tests to verify that the appropriate message is shown when status gets executed.


simonharrer:
Awesome. Thanks for your help. Will be part of the next release. Some lines are wrongly indented, but I'll handle the code formatting myself. Thanks again for your help! :-)

PR #46, opened 2020-06-12
fixed user flag for non-windows systems

enolive:
There is a slight error in the installer that makes it impossible to install mob into user space. I fixed it here


simonharrer:
Thanks!

PR #60, opened 2020-06-25
Scoop installer

ehei:
Created a scoop "app manifest" file which will allow Scoop to install mob on Windows. There is more that could be done for more in depth scoop support for versioning , etc, but it's a step towards that - and it works.

Scoop has the concept of buckets, and you can make your own or contribute to them, so this could be something to add to the 'extras' bucket, or possibly create a new one for mobbing / remote work / etc., which would remove the need to get the mob.json file manually.

Also updated the readme and provided a link to the Scoop site for more information.

The link is hard-coded to the 64 bit windows version - there is support to add 32/64 sections if that seems valuable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ curl -sL install.mob.sh | sh
 ```
 If you are under Windows, you can use `git bash` to install `mob`.
 
-Alternatively, if you use [Scoop](https://scoop.netlify.app/) (alernative to chocolatey, "brew for Windows") then just download the mob.json file, open a PowerShell terminal in that directory and:
+If you use [Scoop](https://scoop.netlify.app/) (alernative to chocolatey, "brew for Windows") then just download the mob.json file, open a PowerShell terminal in that directory and:
 ```
 scoop install mob
 ```
--- a/README.md
+++ b/README.md
@@ -16,7 +16,13 @@ Swift [git handover](https://www.remotemobprogramming.org/#git-handover) with 'm
 ```
 curl -sL install.mob.sh | sh
 ```
-If you are under Windows, you should use `git bash` to install `mob`.
+If you are under Windows, you can use `git bash` to install `mob`.
+
+Alternatively, if you use [Scoop](https://scoop.netlify.app/) (alernative to chocolatey, "brew for Windows") then just download the mob.json file, open a PowerShell terminal in that directory and:
+```
+scoop install mob
+```
+
 
 You can also install it on macOS via homebrew: 
 
--- a/README.md
+++ b/README.md
@@ -16,13 +16,7 @@ Swift [git handover](https://www.remotemobprogramming.org/#git-handover) with 'm
 ```
 curl -sL install.mob.sh | sh
 ```
-If you are under Windows, you can use `git bash` to install `mob`.
-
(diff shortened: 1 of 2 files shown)


mob.json
```diff
@@ -0,0 +1,10 @@
+{
+  "version": "0.0.24",
+  "homepage": "https://github.com/remotemobprogramming/mob",
+  "checkver": "github",
+  "url": "https://github.com/remotemobprogramming/mob/releases/latest",
```
simonharrer, line 5: url should probably be https://github.com/remotemobprogramming/mob/releases/tag/v0.0.24

README.md
```diff
@@ -16,7 +16,13 @@ Swift [git handover](https://www.remotemobprogramming.org/#git-handover) with 'm
 ```
 curl -sL install.mob.sh | sh
 ```
-If you are under Windows, you should use `git bash` to install `mob`.
+If you are under Windows, you can use `git bash` to install `mob`.
+
+If you use [Scoop](https://scoop.netlify.app/) (alernative to chocolatey, "brew for Windows") then just download the mob.json file, open a PowerShell terminal in that directory and:
+```
+scoop install mob
+```
```
simonharrer, line 24: I want to keep the install text at the top of the README short. Can you provide a section below and we link to it?

simonharrer:
Thanks for the PR. I love making 'mob' easy to install. 

For homebrew, there's a separate repo https://github.com/remotemobprogramming/homebrew-brew and because of this repo it's a one liner to install 'mob' via the CLI. Is this possible with scoop as well?

mob.json
ehei: That was the first URL I was using, but it should be able to handle "hey, there's a new version, do you want to update" type of thing pointing at latest. But we can certainly change it to start with just pointing at 24.

README.md
ehei: Sure thing

ehei:
> Thanks for the PR. I love making 'mob' easy to install.
> 
> For homebrew, there's a separate repo https://github.com/remotemobprogramming/homebrew-brew and because of this repo it's a one liner to install 'mob' via the CLI. Is this possible with scoop as well?

Yeah didn't realize there was a separate repo. Can certainly take that route with scoop as well if that's preferred.

simonharrer:
> > Thanks for the PR. I love making 'mob' easy to install.
> > For homebrew, there's a separate repo https://github.com/remotemobprogramming/homebrew-brew and because of this repo it's a one liner to install 'mob' via the CLI. Is this possible with scoop as well?
> 
> Yeah didn't realize there was a separate repo. Can certainly take that route with scoop as well if that's preferred.

A second repo only makes sense if it would enable something like `brew install remotemobprogramming/brew/mob`with scoop. Otherwise, it's not that helpful.

ehei:
> > > Thanks for the PR. I love making 'mob' easy to install.
> > > For homebrew, there's a separate repo https://github.com/remotemobprogramming/homebrew-brew and because of this repo it's a one liner to install 'mob' via the CLI. Is this possible with scoop as well?
> > 
> > 
> > Yeah didn't realize there was a separate repo. Can certainly take that route with scoop as well if that's preferred.
> 
> A second repo only makes sense if it would enable something like `brew install remotemobprogramming/brew/mob`with scoop. Otherwise, it's not that helpful.

Got it...I'll work on cleaning up the README and make the URL change.

simonharrer:
Awesome! Thanks for making mob easier to install! :-)

PR #62, opened 2020-06-27
Add assertions for mob-session branch with qualifier

damienklinnert:
While poking around in this excellent code base, I came across one interesting edge condition that is currently not covered by tests. The behaviour to specify is this;

What shall happen if I am on the branch `mob-session` and specify a branch qualifier of `green`?

* My intuition tells me that `mob-session` should be seen as the `baseBranch` and that `mob/mob-session/green` should be seen as the `wipBranch`. With this solution the function is bijective (there's a 1:1 mapping for all combinations of base and wip branches). 
* Instead, however, the tool reports `master` as the `baseBranch` and `mob/master/green` as the `wipBranch`. This makes the function non-injective (with a qualifier of `green` the branches `mob-session` and `master` both map to `mob/master/green`).

I've attached two assertions for this edge case, but **BEWARE: THE FIRST ASSERTION CURRENTLY FAILS**. To avoid failing builds, don't merge this PR just yet. Think of this PR as a conversation starter instead.

Once we settled on what the behaviour is supposed be, I can change the behaviour accordingly and add it to the pull request.

### DIFF
--- a/mob_test.go
+++ b/mob_test.go
@@ -27,6 +27,9 @@ func TestDetermineBranches(t *testing.T) {
 	assertDetermineBranches(t, "master", "", "", "master", "mob-session")
 	assertDetermineBranches(t, "mob-session", "", "", "master", "mob-session")
 
+	assertDetermineBranches(t, "mob-session", "green", "", "mob-session", "mob/mob-session/green")
+	assertDetermineBranches(t, "mob/mob-session/green", "green", "", "mob-session", "mob/mob-session/green")
+
 	assertDetermineBranches(t, "master", "green", "", "master", "mob/master/green")
 	assertDetermineBranches(t, "mob/master/green", "", "", "master", "mob/master/green")
 


simonharrer:
Thanks for this PR. It's really well written and is a great way to get a conversion going. :-) So here's my part of that conversation. 

'mob' knows two types of branches: base branches and wip branches. Every wip branch originated from a base branch, but there can be multiple wip branches per base branch using the `--branch green` qualifier. 'mob' controls the wip branches. Making this recursive (start a new wip branch of an existing wip branch) is probably not that helpful and adds complexity. That's why being on `mob-session` (a wip branch due to backward compatibility reasons) should ignore any `--branch green` option. Because the branch already is a wip branch and everything can be derived from that fact (base branch is `master` and wip branch is `mob-session` and qualifier is ``).

Does this make sense to you?

damienklinnert:
@simonharrer Thanks for reply. It makes perfectly sense when you put it like that (I would argue that it makes a central piece of the domain knowledge of this application). What do you think about:

a) adding a warning or perhaps even an error with an appropriate error message when the qualifier is used with the `mob-session` branch.
b) encode that knowledge with a test

I'm happy to do those changes to the tool. What do you think?

simonharrer:
Go ahead! I tried to state a few of those things in the readme under design considerations. But maybe there’s a better place?

damienklinnert:
@simonharrer excellent. I don't know yet how things could look like, but I'll play with them very soon. Thanks for your counsel and the direction to move into! I'll close this PR and open a new one when ready (as the direction has changed from what I originally thought).

simonharrer:
great. let me know if I can help.

PR #63, opened 2020-06-28
Added badges for contributors and total downloads

### LINKED ISSUE #53: script to determine download numbers using github api
I would like to know how often 'mob' is downloaded.

The github releases api provides download counts. 

A script that would compute this per release and per platform would be interesting.

Morl99:
Fixes #53


simonharrer:
Love it! Thanks for the contribution! :-)

PR #64, opened 2020-06-28
Add CI workflow

### LINKED ISSUE #59: Run tests via GitHub actions
- run tests on each commit and notify if tests fail
- run tests before each release to ensure that only releases with successful tests can be built

Morl99:
This is a first simple try to adding a CI workflow, that runs all the tests. 
A badge is available here https://img.shields.io/github/workflow/status/remotemobprogramming/mob/ci, so this could help to see if the tests on master are currently passing or failing.

I am not sure, if those tests should also be added to the release workflow. I feel that no commit should make it to master, that is not passing the tests. But I am sure this is strongly opinionated, as I am used to working like this, so feel free to make comments about this. 

I am new to GitHub Actions, but so far I could not find a way of adding the tests to the Release workflow without copy and pasting all of the steps.

Fixes #59

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,13 @@
+on: [push]
+name: Test
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '^1.14.2'
+    - run: git config --global user.email "test@test.test"
+    - run: git config --global user.name "mob test git user"
+    - run: go test
--- a/README.md
+++ b/README.md
@@ -2,12 +2,14 @@
 
 ![mob Logo](logo.svg)
 <p align="center">
-    <a href="https://github.com/remotemobprogramming/mob/graphs/contributors" alt="Contributors">
-        <img src="https://img.shields.io/github/contributors/remotemobprogramming/mob" /></a>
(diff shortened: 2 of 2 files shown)


Morl99:
I guess what I am proposing is to protect the master branch, further reading can be found here: https://help.github.com/en/github/administering-a-repository/about-required-status-checks
let me know what you think, and in which direction you want to go

simonharrer:
I would like to keep things simple. I think running tests for every push is good enough as a first step and a badge in the readme to show that. 

Regarding dependencies: there's the "needs" option which allows to add dependencies between jobs. Not sure, however, if this works across yaml files. Having such a dependency for every release is, to me, the second step.

I would refrain from protecting the master branch through technical measures. I agree that commits on master should have successful tests, but I don't want to enforce that.

Morl99:
I think that is a good first step, and I am all for merging this to master. We can revisit this if new options become apparent. We can add the badge to the readme in this MR and then merge it. (I can probably add it later if you want)

Morl99:
Sorry for the mess I am making of this Pull Request. I am not used to GitHub and got bitten by your merge commit, so that my branch was no longer mergable with your master. I thought I would be able to resolve this with GitHub features, but in the end, I resorted back to git cli and a force push ;)

This should be ready to merge now. The badge will only start to work, once it is merged into master, because right now the Workflow does not exist in your repo.

simonharrer:
Awesome! And no worries on that PR history, the result is what counts. :-)

Issue #61, opened 2020-06-25
Have mob next push if there are commits, but no changes


sjdecaro:
### Problem

During a mob session, one of the developers created a commit while in her session. When she executed `mob next`, she had no additional changes after her last commit. Right now, the `mob next` command skips all `git` commands if no changes are found. So when the next person tried to `mob start`, we couldn't find her work.

### Suggestion

As part of the `mob next` logic, check to see if the local branch is ahead of the remote branch. If so, push to remote.

### Reasoning

We started doing some intermediate commits to have a working state of code to fall back on. This feels like a healthy practice to follow if you get a test working or a refactor completed prior to your session being complete.

If this is not a recommended practice for this tool, that might be helpful to include in the documentation.

simonharrer:
Thanks for helping on improving 'mob'. Really appreciate taking your time in writing up your idea. 

The idea of 'mob' is that it completely controls the wip branch. Have to think about whether it makes sense to deviate from that and support other cases as well.

sjdecaro:
That makes sense, and I totally get that perspective. We didn't really notice speed issues with any of the commands we used, so props there. =)

While I don't want git to cause a latency issue with the handover, our (human) handover was slowed down trying to figure out why our changes were absent. That's why I was hoping for some improved clarity about the proper use of the command, at a minimum (even just "avoid using your normal git commands while using mob" ;) ).

simonharrer:
On second thought, one could locally check if `origin/mob-session` and `mob-session` deviate and only if they point to two different commits initiate a `git push`. This would help you and won't interfere in any other use case.

simonharrer:
Manual commits are pushed now since 0.0.25. 

Please reopen if this doesn't suffice for your use case.

PR #73, opened 2020-07-15
Update mob next to push manual commits

ikuo-suyama:
I implemented this Issue #61 because it's so useful for our usage.
Based on your idea, I fixed mob next to push if the current mob branch and remote commit hash are different.

> On second thought, one could locally check if origin/mob-session and mob-session deviate and only if they point to two different commits initiate a git push.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -414,8 +414,11 @@ func next() {
 	currentBaseBranch, currentWipBranch := determineBranches(gitCurrentBranch(), configuration.WipBranchQualifier, gitBranches())
 
 	if isNothingToCommit() {
-		git("push", "--no-verify", configuration.RemoteName, currentWipBranch)
-		sayInfo("nothing was done, so nothing to commit")
+		if hasLocalCommits(currentWipBranch) {
+			git("push", "--no-verify", configuration.RemoteName, currentWipBranch)
+		} else {
+			sayInfo("nothing was done, so nothing to commit")
+		}
 	} else {
 		git("add", "--all")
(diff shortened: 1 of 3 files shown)


mob.go
```diff
@@ -494,6 +498,14 @@ func isNothingToCommit() bool {
 	return len(strings.TrimSpace(output)) == 0
 }
 
+func hasLocalCommits(currentWipBranch string) bool {
```
simonharrer, line 501: can you rename `currentWipBranch` to just `branch` as this method is quite general in nature?

mob_test.go
```diff
@@ -427,6 +427,45 @@ func TestNothingToCommitCreatesNoCommits(t *testing.T) {
...
 
+func TestStartNextPushManualCommits(t *testing.T) {
+	setup(t)
+
+	setWorkingDir("/tmp/mob/local")
+
+	start()
+	createFile(t, "example.txt", "content")
+	git("add", "--all")
+	git("commit", "-m", "\"asdf\"")
```
simonharrer, line 438: You don't need the double quotes for the commit message, just pass in `"asdf"`.

simonharrer:
Awesome. I love the tests accompanying the change! :-)

Can you write a small entry in the CHANGELOG.md for that? Then, I'll merge it.

mob.go
ikuo-suyama: Sure!

ikuo-suyama:
Thank you for confirming that!
I also wrote CHANGELOG.md.

PR #77, opened 2020-07-24
Add TestStartCommitDoneOnFeatureBranch

dsm2005:
Issue: When completing a mob that was started from a feature branch, mob done does not seem to update the feature branch.

This request contains a new, failing test. It does NOT fix the issue.

There is also some refactoring of the tests included in the merge request. I'm happy to separate these concerns if that's preferred.


simonharrer:
Would like to separate the refactorings from the issue/bug.

simonharrer:
mob done doesn't create commits on the base branch. That's currently by design. It only creates commits in the wip branches.

simonharrer:
I think the design choice of not making commits in the nonwip branches should be kept. I think this is a way to establish trust in the tool. And gibes the user control over the things the user cares about - their own branches. That’s why I close this PR.

PR #79, opened 2020-08-19
Continue with next notification command even if the previous failed

neochrome:
Hey, love your tool - we're using it since a week back and it's really helping us!

I found that I didn't want the voice notification, only the text notification.
The current implementation doesn't continue with the notification command unless the previous (voice) has finished with an exit code 0, which it doesn't unless you have it configured properly.

My proposed changes include
- run all the commands, voice & text regardless if the previous exited ok
~~- add an additional env variable `MOB_NOTIFY_COMMAND` to allow specifying the notification command~~
~~- a small refactoring related to how commands are collected and executed cross platform~~

The last two changes are moved to a separate PR #80

What do you think?


simonharrer:
Awesome. Thanks for your great PR. I agree on your first point. Not sure about the second and third - I have to think about that. Could we try to get the changes of your first point in and then discuss on the second and third point separately? For the first point, a CHANGELOG entry is necessary. Otherwise, it seems good as far as I can see.

neochrome:
Np, I'll split the PR :+1:

PR #80, opened 2020-08-20
Support to specify notify command

neochrome:
- add an additional env variable `MOB_NOTIFY_COMMAND` to allow specifying the notification command
- a small refactoring related to how commands are collected and executed cross platform

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -25,6 +25,7 @@ type Configuration struct {
 	RemoteName                        string // override with MOB_REMOTE_NAME environment variable
 	WipCommitMessage                  string // override with MOB_WIP_COMMIT_MESSAGE environment variable
 	VoiceCommand                      string // override with MOB_VOICE_COMMAND environment variable
+	NotifyCommand                     string // override with MOB_NOTIFY_COMMAND environment variable
 	MobNextStay                       bool   // override with MOB_NEXT_STAY environment variable
 	MobStartIncludeUncommittedChanges bool   // override with MOB_START_INCLUDE_UNCOMMITTED_CHANGES variable
 	Debug                             bool   // override with MOB_DEBUG environment variable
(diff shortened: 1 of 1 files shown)


neochrome:
As I mentioned before I'm interested in having the ability to specify which notification command to run. The default on Linux today uses notify-send, but with default urgency level which means the message is cleared rather quickly (and depending on your config might not be especially prominently styled).
By introducing a new env variable to override the notification command one can switch to another command altogether or change the urgency level to e.g critical (which stays until dismissed).
Without this support one can resort to override the voice command, but then you'll get dual notifications and it also feels kinda hacky :)

neochrome:
Any more thoughts on this matter?

simonharrer:
A few comments, otherwise, I think it's fine.

- executeCommands should be something like executeCommandsAsynchronously or executeCommandsInBackgroundProcess.
- using  ...string instead of []string as the argument of executeCommands should make the code more concise.
- when providing a separate voice or notify command, it would be great if one could be able to place a '%s' in the custom command to specify where the message should be placed in the command. And if it's missing, it'll be added at the end of the custom command as it's currently done.

simonharrer:
And thanks for the patience, I'm just back from vacation - that's why it took so long.

neochrome:
Good comments, I'll tweak and update the PR 👍
Also, no worries, vacation is essential  and I'm rocking my branch/fork
locally so I already have the added functionality available 😎

neochrome:
What about something like this? As an additional benefit one is now allowed to override the voice command also on windows if wanted.

simonharrer:
At the moment, the configuration.notifyCommand is empty by default, and the configuration.voiceCommand is set to "say". Both should work similarly, so either both are empty by default or set. Maybe, we should set them to their OS-specific value by default. So one could inspect how they are called and could have a look at the command via 'mob config'. What do you think?

neochrome:
That seems like a good idea! Let me fiddle a bit more :)

neochrome:
What do you think about something like this?

The last commit is just an idea that you could have the option to switch off a command by providing an empty string. Without this change it's not possible since an empty string would fall back to the default config for that command. Does that make sense?

simonharrer:
Looks great. I'll merge this. Thanks for your help! PRs like this makes maintaining an open source project a breeze. :-)

neochrome:
Glad to help out :)

PR #84, opened 2020-11-05
After installing, check if the mob binary is at the expected location

L7R7:
I had a manually installed version on my machine that was preferred over the one that was installed by `install.sh`. I didn't notice that for quite a while, because it wasn't obvious from the scripts' output.

This PR adds a check for that and will echo a warning if the `mob` binary ist not at the expected location. I tested it on my machine (Ubuntu 20.04), not sure if this works on all platforms.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -144,6 +144,15 @@ check_say() {
   esac
 }
 
+check_installation_path() {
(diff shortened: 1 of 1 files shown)


simonharrer:
LGTM

PR #85, opened 2020-11-13
Read WIpBranchQualifier from an environment variable

NeilSorensen:
The old behavior of `MOB_WIP_BRANCH` was important for us, since we have multiple teams collaborating on a single codebase.
While it's technically possible to keep multiple branches going with the `-b` flag, having to specify the same branch every time is tedious.

This adds a new environment variable, `MOB_WIP_BRANCH_QUALIFIER`, that allows you to set a default branch qualifier.
Setting a value on the command line overrides the environment variable, allowing you to specify a different wip branch.
To use the default branch when the `MOB_WIP_BRANCH_QUALIFIER` environment variable is set, you can specify `-b ""`

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	versionNumber   = "0.0.27"
+	versionNumber   = "0.0.26"
 	mobStashName    = "mob-stash-name"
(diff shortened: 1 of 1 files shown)


simonharrer:
Makes sense. Thanks for the help.

PR #86, opened 2020-11-21
How to update using Homebrew

tsundberg:
For some this may be obvious, for some it is just nice to have an example you can copy and paste.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ You can also install it on macOS via homebrew:
 brew install remotemobprogramming/brew/mob
 ```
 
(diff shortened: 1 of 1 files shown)


simonharrer:
Personally I only key in the install command again which issue an update if there’s a newer version. Is the update command the preferred approach?

tsundberg:
I don't know what the preferred approach is :)

I didn't even consider re-running the installation command as I wanted to update the version. Running the install command felt incorrect to me.

simonharrer:
Ok. :-)

PR #87, opened 2020-11-28
Ignore branch qualifier when branch is mob-session

damienklinnert:
This pull request is a follow up to #62. 

It changes the code to match this specification from the previous pull request:

> That's why being on mob-session (a wip branch due to backward compatibility reasons) should ignore any --branch green option.

The changes in this pull request will silently ignore the branch (and not warn the user about what exactly is going wrong). I'm not entirely sure how to add that warning or if we even need one. What do you think @simonharrer ?

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -267,13 +267,9 @@ func determineBranches(branch string, branchQualifier string, branches string) (
 		suffix = preparedBranch[index:]
 	}
 
-	if branch == "mob-session" || branch == "master" {
+	if branch == "master" && branchQualifier == "" || branch == "mob-session" {
 		baseBranch = "master"
-		if branchQualifier != "" {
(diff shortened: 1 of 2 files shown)


simonharrer:
Basically, using `--branch` (or via the environment variable) only makes sense when one is on a base branch. When one is on a mob branch, it can and must be determined automatically. Ignoring it in that case is fine. Can we extend the tests for ignoring it for branches like `mob/main` or `mob/main/red` as well? Especially the latter is interesting when the qualifier differs from the configuration option.

damienklinnert:
Just for future reference: the changes in this pull request have been superseded by the ["Rework determineBranches() implementation" commit](https://github.com/remotemobprogramming/mob/commit/757c1b2c7a491c8417932752136a0bd006eff416).

PR #93, opened 2020-12-22
Improve instructions for how to contribute

jbrains:
When I followed the "How to contribute" instructions as is, I couldn't run the tests. I learned that I need to run the `create-testbed` script, but when I ran that script, it failed. I believe that this PR fixes both problems.


simonharrer:
Thank you very much for making it simpler for new developers to start contributing!

PR #96, opened 2020-12-26
Adds the go.mod file, and further refactoring

gregorriegler:
The go.mod file helps with IntelliJ.
It enables to show code coverage and debug tests.
Maybe you want to change the module naming.

With the help of your tests I did more refactoring.
Was focusing on determineBranches.
I did many smaller commits, so it is more comprehensible.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module remotemobprogramming.org/mob
+module remotemobprogramming.org/mob.sh
 
 go 1.15
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module remotemobprogramming.org/mob
+
+go 1.15
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module remotemobprogramming.org/mob.sh
+module mob.sh
 
 go 1.15
--- a/mob.go
+++ b/mob.go
@@ -258,27 +258,13 @@ func determineBranches(currentBranch string, userSpecifiedBranchQualifier string
 	}
 
 	if currentBranch == "mob-session" || (currentBranch == "master" && userSpecifiedBranchQualifier == "") {
-		baseBranch = "master"
 		wipBranch = "mob-session"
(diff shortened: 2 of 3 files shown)


simonharrer:
Can you use mob.sh as the module name?

I like the say refactorings.

 I’m not sure I like the refactorings for the determineBranches. Easiest thing would be to split this PR so the module file and the say refactorings can be merged and we can discuss the determineBranches in another PR. 

A little bit of background already: the determineBranches method is very central to the tool. I want to keep it small and concise so one can understand the relationship between base and wip branch at a glance. That’s why I want to have the business logic within a single method with two return values and generic helper methods. @damienklinnert as we refactored it a few weeks ago, what do you think?

gregorriegler:
> Can you use mob.sh as the module name?

Yes, that worked.

> I’m not sure I like the refactorings for the determineBranches. Easiest thing would be to split this PR so the module file and the say refactorings can be merged and we can discuss the determineBranches in another PR.
A little bit of background already: the determineBranches method is very central to the tool. I want to keep it small and concise so one can understand the relationship between base and wip branch at a glance. That’s why I want to have the business logic within a single method with two return values and generic helper methods. 

Ok. For me it was smelly, and it was confusing me. I could not wrap my head around it until I refactored it.
The helper functions did a lot of work in this regard. If you want to join it again, I think the easiest way would be to "refactor join" it again. Then the helper functions would stay, and the arguments are still fixed.
What do you say?

go.mod
```diff
@@ -0,0 +1,3 @@
+module remotemobprogramming.org/mob.sh
```
simonharrer, line 1: I meant having only `module mob.sh`

simonharrer:
Hm, I think doing a quick pairing session would really help here. If you have some time spontaneously today or in the next few days, we could get this done quickly.

gregorriegler:
ok. i'm on a mobile internet right now, but we can try. i had enough bandwith on the speedtest. write me on twitter or so where i can join you

Issue #101, opened 2021-01-01
os package (brew, pacman, apt, ...)?


draptik:
Is there interest in making OS packages for mob?

For example:

- Arch: pacman -S mob
- Debian: apt install mob
- Windows: chocolatey install mob
- Mac: brew install mob
- etc

Has anybody looked into this? Are there problems?
As far as I can see the only dependencies are go and git. Am I missing something?

simonharrer:
Regarding brew: I think the current solution is fine as we do have brew but instead of delivering source code our brew repository delivers a binary. (official brew apps must always be built from source code locally).

Regarding the others: feel free to help. :-)

simonharrer:
We do have something for scoop as well. But it’s not published in official repos.

draptik:
Here is an Arch AUR package, using the binary provided by https://github.com/remotemobprogramming/mob/releases/.

https://aur.archlinux.org/packages/mobsh-bin/

Note: I couldn't use the name `mob` because another package is already using the name. Therefore `mobsh`...

simonharrer:
Awesome. Would you like to add install instructions to the README of mob? And if you like, you can setup a guthub action so the archlinux repo will be updated automatically.

draptik:
I've added a second arch package which compiles the sources from scratch (and runs the tests):

https://aur.archlinux.org/packages/mobsh/

It uses slightly different Go compile flags as recommended by the arch wiki: https://wiki.archlinux.org/index.php/Go_package_guidelines#Flags_and_build_options

draptik:
Concerning Github Actions: I'll have to look into this. I might have to add you (or some other maintainer) as Co-Contributor for the AUR package (ssh-keys etc)...

simonharrer:
Alternatively I could add you on the mob repo as a collaborator. Just let me know if this would help.

simonharrer:
Regarding whether shipping binaries or building from source: you probably know best which is the best way for archlinux users.

draptik:
Github Actions: I would like to first manually go through the version upgrade process for the Arch package a couple of times so I can get a feel for the workflow. I'll get back to you once I have a better understanding of the steps involved.

simonharrer:
Should we close the issue for now?

draptik:
Yes, the issue can be closed.

PR #103, opened 2021-01-04
README: add arch package installation instructions

draptik:
See https://github.com/remotemobprogramming/mob/issues/101 for details.


simonharrer:
Thanks for your help!

PR #106, opened 2021-01-06
Found a way to parameterize the Environment Variable Tests

gregorriegler:
Uses [SubTests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) to parameterize all expectations for boolean typed environment variables. To test all cases for a single boolean environment variable all it needs is a single line of code.

It indents the cases nicely
image
Also it does not fail all Tests.
In the following example i have only implemented the default of MOB_DONE_SQUASH wrong
image

### DIFF
--- a/mob_test.go
+++ b/mob_test.go
@@ -102,28 +102,31 @@ func TestMobDoneSquashEnvironmentVariableDefault(t *testing.T) {
 
 func TestMobDoneSquashEnvironmentVariableEmpty(t *testing.T) {
 	// ASSUME default setting for mobDoneSquash is true
-	configuration = setEnv("MOB_DONE_SQUASH", "")
-
-	equals(t, true, configuration.MobDoneSquash)
+	parameterizedEnvVarTest(t, "MOB_DONE_SQUASH", "", true, func(configuration Configuration) interface{} { return configuration.MobDoneSquash })
 }
 
 func TestMobDoneSquashEnvironmentVariableTrue(t *testing.T) {
-	configuration = setEnv("MOB_DONE_SQUASH", "true")
-
-	equals(t, true, configuration.MobDoneSquash)
+	parameterizedEnvVarTest(t, "MOB_DONE_SQUASH", "true", true, func(configuration Configuration) interface{} { return configuration.MobDoneSquash })
 }
 
 func TestMobDoneSquashEnvironmentVariableFalse(t *testing.T) {
-	configuration = setEnv("MOB_DONE_SQUASH", "false")
-
-	equals(t, false, configuration.MobDoneSquash)
+	parameterizedEnvVarTest(t, "MOB_DONE_SQUASH", "false", false, func(configuration Configuration) interface{} { return configuration.MobDoneSquash })
 }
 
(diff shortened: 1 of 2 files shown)


mob_test.go
```diff
@@ -76,70 +76,50 @@ func assertDetermineBranches(t *testing.T, branch string, qualifier string, bran
...
-func TestMobDoneSquashEnvironmentVariableEmpty(t *testing.T) {
-	// ASSUME default setting for mobDoneSquash is true
-	os.Setenv("MOB_DONE_SQUASH", "")
-	defer os.Unsetenv("MOB_DONE_SQUASH")
-
-	configuration = parseEnvironmentVariables(getDefaultConfiguration())
-
-	equals(t, true, configuration.MobDoneSquash)
+func TestBooleanEnvironmentVariables(t *testing.T) {
+	expectEnvironmentVariableSetsBooleanConfiguration(t, "MOB_DONE_SQUASH", true, func(configuration Configuration) interface{} { return configuration.MobDoneSquash })
```
simonharrer, line 92: One can write such functions as part of getters for that struct. 

```golang
func (c Configuration) GetMobDoneSquash() bool {
    return c.MobDoneSquash
}
```

and with this function the test should become more readable, as we only need to pass in that function as a parameter.

simonharrer: Here's the article from which I got my info: https://medium.com/rungo/anatomy-of-methods-in-go-f552aaa8ac4a

mob_test.go
```diff
@@ -76,70 +76,50 @@ func assertDetermineBranches(t *testing.T, branch string, qualifier string, bran
...
+	})
 }
 
-func TestMobDoneSquashEnvironmentVariableGarbage(t *testing.T) {
-	// ASSUME default setting for mobDoneSquash is true
-	os.Setenv("MOB_DONE_SQUASH", "garbage")
-	defer os.Unsetenv("MOB_DONE_SQUASH")
+func setEnv(variable string, value string) Configuration {
+	os.Setenv(variable, value)
+	defer os.Unsetenv(variable)
```
simonharrer, line 120: I think I would rename the function to something like `setAndParseEnvironmentVariableInConfiguration` to make it really clear what happens here. What do you think?

simonharrer:
I really like the visualization of the tests in IntelliJ. Great contribution. Just had a few minor remarks.

simonharrer:
Another remark: I don't really know whether the (=true) in the IntelliJ output is the actual or the expected value. Maybe we should make this more clear.

gregorriegler:
I made the changes you suggested. Please read through the commit messages which are a bit longer this time, as there are important things to note.

simonharrer:
I meant something like this https://stackoverflow.com/questions/38897529/pass-method-argument-to-function where one still use lambdas but pass in the existing method on the struct

gregorriegler:
You were right. It works nicely by passing the getters as Method Expressions.
Just had to convert the return type from bool to interface{} using a higher-order function.
I think its much better now.

simonharrer:
Thanks for your help!

PR #107, opened 2021-01-07
add option to require a commit message when running`mob next` 

stevegt:
Hi All,

One of the first things we found we wanted when starting to use `mob` was a way to remind us to provide the `-m` flag when running `mob next`.  This pull request provides that feature, by implementing an optional MOB_REQUIRE_COMMIT_MESSAGE environment variable that defaults to false. When true, the feature requires the `-m` flag.

(In our case, we want to capture those intermediate progress messages and turn most of them into bullet points when we later run `mob done`.)

- add MOB_REQUIRE_COMMIT_MESSAGE bool env variable
  - defaults to false so current behavior is unchanged
- add MOB_REQUIRE_COMMIT_MESSAGE to `mob config` output

Co-authored-by: jessicajt <jessica@t7a.org>
Co-authored-by: Ryan-pelo <xfactor529@gmail.com>
Co-authored-by: angela <angela@t7a.org>
Co-authored-by: koshchei <m@nordling.org>


gregorriegler (commented):
Some considerations on the Code:

- The logic checking whether a message was given is at the wrong place. Belongs in `next`
- There are no tests for this. All the other code is tested, as this feature would have to.

simonharrer:
First of all, thanks for helping to make the mob tool even better. And thanks for taking your time to open this PR. I'm okay with this feature as it is to be part of your workflow.

Regarding "The logic checking whether a message was given is at the wrong place. Belongs in next": I would add another variable in the configuration similar to MobNextStaySet which is set during parsing and can later be evaluated in the `next()` method. 

Regarding "There are no tests for this. All the other code is tested, as this feature would have to.": I agree. There are plenty of tests that can act as an example for testing this behavior. Please have a look at `mob_test.go` and feel free to reach out in this PR.

stevegt:
> Regarding "The logic checking whether a message was given is at the wrong place. Belongs in next": I would add another variable in the configuration similar to MobNextStaySet which is set during parsing and can later be evaluated in the `next()` method.

That's the best alternative we could think of.  We weren't sure you'd want the Configuration struct modified, but we also didn't want to modify the way returns and parameters work between parseArgs() and next().  Okay, we'll add MobNextStaySet and move the logic to next().  We should be able to get to that in one of our sessions this week.

> Regarding "There are no tests for this. All the other code is tested, as this feature would have to.": I agree. There are plenty of tests that can act as an example for testing this behavior. Please have a look at `mob_test.go` and feel free to reach out in this PR.

Simon, do you have a rule of thumb for how much or what sort of coverage you'd want, something you'd be comfortable having in the "How to contribute" section of the README?  I like to get close to 100% myself, mocking as needed, but right now mob.go is only around 70.9%, with some pretty large swaths uncovered.  We had a discussion that night and decided to submit this PR without tests in part because of the latter statistic (it was late).  

Either way, this is a helpful nudge toward better TDD -- @gregorriegler's comment caused me to change the way we've been running our sessions; since last week I've been kicking things off by writing a failing test first and letting them figure things out. At least this way they're starting to get more experience writing tests by fixing the logic errors in mine.  ;-)

gregorriegler:
Hi @stevegt, I am happy that you're sticking with the pull request.

I would add one test that has `MOB_REQUIRE_COMMIT_MESSAGE` set to `true`, and does an arbitrary file change followed by a `mob next` without a message.
This test would verify, that the `mob next` was rejected, and no further commit occurred. 
I would only check the status of the git log for this, and dismiss the console output. There are helper functions to achieve this in `mob_test.go`

Then I would aim to have another test that checks the same scenario but with a `mob next -m "message"`.
This one would verify, that the `mob next` command succeeded, and the wip commit is present in the git log.

Finally, I would aim to have another test that checks the same scenario, but with `MOB_REQUIRE_COMMIT_MESSAGE` set to `false`. This Test would verify, that the `mob next` command succeeded as the commit is found in the git log.
Possibly, we do already have a test case for this final scenario. In that case, I would just make sure that it already has the proper assertions.

Other than that, I would not mind the code coverage.

simonharrer:
No worries regarding submitting the PR without tests. It got you a PR we could discuss on. That was the perfect way. :-)

Regarding test coverage: I'm not a fan of 100% coverage. I'm a fan of testing the things I am uncertain about. Mostly leads to coverage rates between 60% and 90%. It may be also because I'm not a fan of mocks which, in my mind, make larger refactorings hard - instead, I like to have tests on the outside and be able to refactor in any way in the inside without having to rewrite a bunch of tests. Hope this clears up the philosophy currently behind the test strategy of the tool.

stevegt:
This is mostly done; it's late in the evening when we're pushing this, so I would not merge yet.  Coverage is good, though there might be another corner case or two I'd want to look for when fresh.  One thing we did just now notice is that there should be some logic to not require the commit message if there are no local changes.  

We wound up not needing the *Set field in configuration; we did need to compare the default commit message with the message provided with `-m` in order to detect the `-m`.

simonharrer:
With your current change, the following three cases will succeed. I, however, expected the second case to fail as the failure message in your PR is "sayError("-m 'commit message' required")". So either we fix the error message or we fix the behavior.

```bash
# case 1
export MOB_WIP_COMMIT_MESSAGE="my very own wip commit message"
export MOB_REQUIRE_COMMIT_MESSAGE=false

mob start
mob next
```

```bash
# case 2
export MOB_WIP_COMMIT_MESSAGE="my very own wip commit message"
export MOB_REQUIRE_COMMIT_MESSAGE=true

mob start
mob next
```

```bash
# case 3
export MOB_WIP_COMMIT_MESSAGE="my very own wip commit message"
export MOB_REQUIRE_COMMIT_MESSAGE=true

mob start
mob next --message "custom message"
```

stevegt:
Thank you Simon.  Case 2 is one of those corner cases I was thinking about last night.  I think we likely have the behavior right, and the failure message wrong.  In other words, if the user provides `MOB_WIP_COMMIT_MESSAGE`, then that should satisfy the `MOB_REQUIRE_COMMIT_MESSAGE=true` requirement, so I think the right thing to do here is `sayError("commit message required")`.

I want to keep exercising our mob's workflow, so rather than fix this myself I'll change the test case on our mob/require-commit-message branch; that will trigger a `make test` failure in our internal monorepo and should result in a new commit showing up here in a few days.

mob_test.go
```diff
@@ -166,6 +166,31 @@ func TestNextNotMobProgramming(t *testing.T) {
...
+	configuration = parseEnvironmentVariables(getDefaultConfiguration())
+	equals(t, false, configuration.RequireCommitMessage)
+
+	os.Setenv("MOB_REQUIRE_COMMIT_MESSAGE", "false")
+	configuration = parseEnvironmentVariables(getDefaultConfiguration())
+	equals(t, false, configuration.RequireCommitMessage)
+
+	os.Setenv("MOB_REQUIRE_COMMIT_MESSAGE", "true")
+	configuration = parseEnvironmentVariables(getDefaultConfiguration())
+	equals(t, true, configuration.RequireCommitMessage)
```
simonharrer, line 184: For boolean environment variables, there's a parameterized test available: ` TestBooleanEnvironmentVariables`. You can simply add your variable `MOB_REQUIRE_COMMIT_MESSAGE` and it will be covered by four tests. And you can simplify the test above as you just set the `configuration.mobRequireCommitMessage = true` without any environment variables.

angelajt:
We've fixed the error so that it just says "commit message required", and handled the case when there's nothing to commit.  Still need to integrate @simonharrer's recommendation to use `TestBooleanEnvironmentVariables`.

simonharrer:
Awesome. Looking forward to get this PR merged! I'll plan to quickly release 1.2.0 afterwards.

angelajt:
This looks done as far as we can tell. Sorry for the delay; we had several other things to deal with.

simonharrer:
No worries. Thanks for your help! :-)

PR #112, opened 2021-01-23
Encapsulate Configuration

gregorriegler:
Hi @simonharrer 

I continued the refactoring. Now that we are using the getter for MobDoneSquash I thought it would make sense to only use it.
And that is the first commit.

Then I was thinking about the writes to this config boolean, and to encapsulate it with a Setter.
I'm not a big fan of get/set for properties, but I'm also not a fan of having both ways (with accessor, and without).
And after all, the getter enabled neat testing of the config.

The Setter was a bit trickier, as I had to do a couple of steps and solve a couple of problems.
I had to do a split-phase refactoring in the setBoolFromEnv function and extract a parseBoolFromEnv function that returns an optional boolean.
I managed to represent an optional boolean with a pointer however, I don't like the code yet.
I think an OptionalBoolean type or an Optional monad would be better.
So you see, it's getting a little complicated.
So this is kind of experimental. But still, I think the code could become more expressive in the end.
Let me know how you think about it.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -137,36 +137,17 @@ func setOptionalStringFromEnvVariable(s *string, key string) {
 func setBoolFromEnvWithSetter(setter func(value bool), key string) {
 	env := parseBoolEnv(key)
 	if env != nil {
+		debugInfo("overriding " + key + " =" + strconv.FormatBool(*env))
 		setter(*env)
 	}
 }
 
 func setBoolFromEnvVariable(s *bool, key string) {
 	optionalBool := parseBoolEnv(key)
-
 	if optionalBool != nil {
+		debugInfo("overriding " + key + " =" + strconv.FormatBool(*optionalBool))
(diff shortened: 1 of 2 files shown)


gregorriegler:
Just realized that you added MobNextStaySet.
It is currently not used, but I think I get the Idea.
It seems you see value in knowing whether a configuration property was set by the user or not after the property is "locked-in".
The pattern is similar with the os.getEnv function. Is this something you want to add to the other properties too?
If so, I'm thinking whether it would be feasible to encapsulate these two details (set bool and actual value) in its own type.

simonharrer:
Not sure this is the right way. Those setter and getter feel wrong and boilerplate.

gregorriegler:
We could keep the getter as Testcode only.
Means i move it from mob.go to mob_test.go
Then mob.go uses the properties directly.

PR #116, opened 2021-01-26
Add '--iuc' param

MikeVitale42:
Added '--iuc' command-line param as an alias for '-i'/'--include-uncommitted-changes'

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -183,7 +183,7 @@ func parseArgs(args []string) (command string, parameters []string) {
 	for i := 1; i < len(args); i++ {
 		arg := args[i]
 		switch arg {
-		case "--include-uncommitted-changes", "-i":
+		case "--include-uncommitted-changes", "--iuc", "-i":
 			configuration.MobStartIncludeUncommittedChanges = true
 		case "--debug":
 			configuration.Debug = true
@@ -433,6 +433,7 @@ func start() {
 			sayUnstagedChangesInfo()
 			sayUntrackedFilesInfo()
 			sayTodo("To start mob programming including uncommitted changes, use", "mob start --include-uncommitted-changes")
+			sayTodo("Or", "mob start --iuc")
 			return
 		}
 	}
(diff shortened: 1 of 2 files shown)


MikeVitale42:
The command-line parameter '--include-uncommitted-changes' is very descriptive, but rather long. It has led to my teammates choosing to use the mouse to highlight it to make sure they spelled everything correctly. This patch adds '--iuc' as a parameter alias.

simonharrer:
Thanks for your PR. 

To be honest, however, I am not sure how to proceed. When mob encounters a situation where it cannot solve the situation by itself, it will tell the user **one** command with which the situation could be fixed. That command should be easy to understand without any abbreviations and easy to copy and paste so there are no typo errors which may lead to other issues.

That being said, I have no issue in adding --iuc, but I am against adding an additional todo message (or an abbreviated one) when the user executes `mob start` with uncommitted changes present.

My hypothesis: when one works with the tool for some time, the `-i` option would be preferred, even over the `--iuc` one.

gregorriegler:
There is already a long and a short version of that param.
Another thing to consider is the complexity of the codebase. Do we really want more conditions to maintain for a feature that adds very little, or arguably no additional value?

simonharrer:
Having thought about it the last two days, I think the `-i` is the intended way to go here. That's why I'm closing this PR. Feel free to reopen if you want to get a discussion going.

MikeVitale42:
Understood on closing an unrequested PR for functionality that already exists.

I fully understand the fully-descriptive name of `--include-uncommitted-changes`, but I still have a hard time typing that out myself (and I'm pretty quick with typing.)  Some other members of my mobs are not always as fast, and for many, English is not their first language.

Perhaps an alternate thing to do would be to have the error message in this case be the short version of the parameter instead of the long one?

Either way, I'll drop it after this unless further discussion is warranted.  I still got my first experience with the Go language.  :)

simonharrer:
I think the long name is great for people that use the tool for the very first time. My assumption is that when you use the tool more often, at some point you have a look at the options (or at least one in your team) and you'll learn about the `-i`flag that saves you the typing.

PR #121, opened 2021-02-01
Automatically append `Co-authored-by:` lines after squash

### LINKED ISSUE #81: Automatically Populate Co-Authored-By lines
Because each of the WIP commits is generated by the authors in the group, it is trivial to aggregate the WIP commits and generate a unique list of Co-Author lines. I have written a little wrapper script in bash that does this when I call the done command.

```bash
CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
FROM_BRANCH=master
if [[ "$CURRENT_BRANCH" != "mob-session" ]]
then
  FROM_BRANCH=`echo "$CURRENT_BRANCH" | awk -F/ '{print $2}'`
fi

CO_AUTHOR_LINES=$(git log $FROM_BRANCH.. --pretty=format:"Co-Authored-By: %an <%ae>" | sort | uniq | grep -v `git config user.email`)

mob done

printf "\nRecommended Co-Author Lines have been copied to your clipboard:\n\n"
printf "$CO_AUTHOR_LINES"
printf "\n\n\n$CO_AUTHOR_LINES" | pbcopy
```

I think it would be cool to integrate this into mob done command. I am possibly willing to submit a PR but wanted to get a sense of the maintainers' opinion on this before I put in the effort.

Some things to consider:
- My implementation is not very robust at computing branch names. It is sufficient for my needs. If integrated into the mob done command, then of course this would be improved.
- Right now all I am doing is copying it to the user's clipboard. I'm assuming OSX and therefore using pbcopy. If integrated into mob done, I would need a cross-platform solution for this.
- Ideally, I would love to auto-populate this into a template so it just automatically shows up in the user's editor when they do their next commit. I haven't figured out a way to do this though. Maybe copying to the clipboard is the best we can do, but I'm curious if any git masters out there know a way to make this more seamless.

stevegt:
In draft mode for discussion:

- fixes #81
- includes a TestStartDoneCoAuthors() test case
- includes a refactored create-testbed that adds more local repos and sets unique committers in each to exercise coauthors code

Implementation in mob.go will need (approximately):

- an appendCoAuthors() function, called by done(), modeled after the bash code in https://github.com/remotemobprogramming/mob/issues/81#issuecomment-770440672 
  - include a "Co-authored-by: Dustin McQuay <dmcquay@gmail.com>" header in this commit
- a gitLogCoAuthors() function, modeled after gitBranches(), called by appendCoAuthors().  We would call the following `git log` command to get the list of authors on the mob branch, splitting each resulting line on comma to get the username and email.  We would also need to remove duplicates from the resulting list, and would need to remove the user who is running `mob done` before returning the list to appendCoAuthors().
```
runCommand("git",  "log", fmt.Sprintf("%s..", currentBaseBranch), "--reverse", "--pretty=format:%an,%ae")
```
- a gitUserEmail() function, modeled after gitUserName(), both called by gitLogCoAuthors(), so we can remove the current user from the coauthors list

I'll recommend that we write a separate test case for each of the above, starting from the bottom, before implementing each.

If there are no objections, I'd like to change the `/tmp/mob` basedir used for testing. To reduce risk, this should either be in e.g. a ./var directory that is in .gitignore, or if it must for some reason be in /tmp then we ought to use a `tempfile`-style algorithm to prevent collisions.  An idiomatic method would be to call t.TempDir() and pass that path to create-testbed.

If create-testbed needs to be touched much more, it may make sense to port it to Go and move it into mob_test.go -- that would also get rid of the duplication of hardcoded paths between those two files.  I'm thinking we should try to avoid that extra work for this PR though.

We'll work this locally when it comes up in our mobbing queue, and I'll also keep an eye on pull requests against https://github.com/t7a/mob/tree/append-coauthors in case anyone else wants to add commits.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -648,7 +648,7 @@ func getCachedChanges() string {
 func makeWipCommit() {
 	commitMsg := configuration.WipCommitMessage
 	git("add", "--all")
-	git("commit", "--allow-empty", "--message", commitMsg, "--no-verify")
+	git("commit", "--message", commitMsg, "--no-verify")
 }
 
 func done(configuration Configuration) {
--- a/mob.go
+++ b/mob.go
@@ -624,8 +624,8 @@ func next(configuration Configuration) {
 			sayInfo("nothing was done, so nothing to commit")
 		}
 	} else {
-		git("add", "--all")
-		git("commit", "--message", configuration.WipCommitMessage, "--no-verify")
+		makeWipCommit()
+
 		changes := getChangesOfLastCommit()
 		git("push", "--no-verify", configuration.RemoteName, currentWipBranch)
 		say(changes)
@@ -645,6 +645,12 @@ func getCachedChanges() string {
 	return strings.TrimSpace(silentgit("diff", "--cached", "--stat"))
 }
 
+func makeWipCommit() {
+	commitMsg := configuration.WipCommitMessage
+	git("add", "--all")
+	git("commit", "--allow-empty", "--message", commitMsg, "--no-verify")
+}
+
 func done(configuration Configuration) {
 	if !isMobProgramming(configuration) {
 		sayError("you aren't mob programming")
@@ -658,8 +664,7 @@ func done(configuration Configuration) {
 
 	if hasRemoteBranch(currentWipBranch, configuration) {
 		if !isNothingToCommit() {
-			git("add", "--all")
-			git("commit", "--message", configuration.WipCommitMessage, "--no-verify")
+			makeWipCommit()
 		}
 		git("push", "--no-verify", configuration.RemoteName, currentWipBranch)
 
@@ -674,6 +679,10 @@ func done(configuration Configuration) {
 		git("push", "--no-verify", configuration.RemoteName, "--delete", currentWipBranch)
 
 		say(getCachedChanges())
+		err := appendCoauthorsToSquashMsg(workingDir)
+		if err != nil {
+			sayError(err.Error())
+		}
 		sayTodo("To finish, use", "git commit")
 	} else {
 		git("checkout", currentBaseBranch)
@@ -914,6 +923,7 @@ func gitignorefailure(args ...string) error {
 	}
 	return err
 }
+
 func runCommand(name string, args ...string) (string, string, error) {
 	command := exec.Command(name, args...)
 	if len(workingDir) > 0 {
--- a/mob.go
+++ b/mob.go
@@ -797,6 +797,10 @@ func gitUserName() string {
 	return strings.TrimSpace(silentgit("config", "--get", "user.name"))
 }
 
+func gitUserEmail() string {
+	return strings.TrimSpace(silentgit("config", "--get", "user.email"))
+}
+
 func showNext(configuration Configuration) {
 	debugInfo("determining next person based on previous changes")
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - The list of commits included in a mob branch is now truncated to a maximum of 5 entries to prevent the need for scrolling up in order to see the latest included changes.
 - Show more informative error message when `mob <cmd>` is run outside of a git repository.
 - Add environment variable MOB_TIMER which allows setting a default timer duration for `mob start` and `mob timer` commands.
+- Add automatic co-author attribution. Mob will collect all committers from a WIP branch and add them as co-authors in the final WIP commit.
 
 # 1.3.0
 - The default `MOB_COMMIT_MESSAGE` now includes `[ci skip]` and `[skip ci]` so that all the typical CI systems (including Azure DevOps) will skip this commit.
--- a/coauthors.go
+++ b/coauthors.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+// Author is a coauthor "Full Name <email>"
+type Author = string
+
+func collectCoauthorsFromWipCommits(file *os.File) []Author {
+	var committer string
+	coauthorsHashSet := make(map[Author]bool)
+
+	authorOrCoauthorMatcher := regexp.MustCompile("(?i).*(author)+.+<+.*>+")
+	scanner := bufio.NewScanner(file)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if authorOrCoauthorMatcher.MatchString(line) {
+			author := stripToAuthor(line)
+
+			// committer of this commit should
+			// not be included as a co-author
+			if committer == "" || author == committer {
+				committer = author
+				continue
+			}
+			coauthorsHashSet[author] = true
+		}
+	}
+
+	coauthors := make([]string, 0, len(coauthorsHashSet))
+
+	for k := range coauthorsHashSet {
+		coauthors = append(coauthors, k)
+	}
+	sort.Sort(byLength(coauthors))
+
+	return coauthors
+}
+
+func appendCoauthorsToSquashMsg(workingDir string) error {
+	squashMsgPath := path.Join(workingDir, ".git", "SQUASH_MSG")
+	file, err := os.OpenFile(squashMsgPath, os.O_APPEND|os.O_RDWR, 0644)
+	if err != nil {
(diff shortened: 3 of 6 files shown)


simonharrer:
First of all, thanks for helping out with #81 and investing your time making the mob tool better.

Reading your great summary above, a few things came to my mind. I would value your opinion on them.

1. Why is the user issueing `mob done` excluded? I was thinking in the lines of simplifying the implementation by having less exceptions and rules. And from my experience in my team, that user would be part of the work anyways.
2. Why do you need to modify the test bed? I might have missed something, but I currently cannot think of a reason why two local repositories and one remote repo should not suffice to check whether this feature works. The thing is, the test bed is executed quite a lot and keeps the tests runnning quickly. Setting unique committers would also be done within the tests itself right before a `mob next`. So this would keep the test bed simple and the important stuff would be part of the actual test.

stevegt:
Hi Simon,

1. The user issuing `mob done` shows in the Author: header on the resulting commit, so also adding them to the Co-authored-by: headers is redundant and possibly confusing to people and tools.  Convention seems to agree with leaving them out, and you'll notice that @dmcquay and I both filtered that the same way in our bash implementations.  (I purposely came up with a first implementation independently before examining Dustin's code, just to see if I would discover anything he missed.  The only major difference was in how we discovered authors -- I parsed SQUASH_MSG for them, while he used git porcelain.  I went with his method in the version at https://github.com/remotemobprogramming/mob/issues/81#issuecomment-770440672.)
2. Four local repos, each configured as a different git user, is more representative of a realistic mob.   Two local repos isn't enough; that only gives us a single co-author, so we can't check several conditions such as sorting and formatting of coauthor lines.  When evolving the bash code, I found I needed at least three repos to test, and I found that four repos makes it easier to reason about how tests should work in regards to sorting and deduplication.  You'll notice that the new create-testbed, with unique authors added, passes with no modifications to any existing tests.  On my system it adds about one second of total `go test` time for each additional repo created. We could reduce that by passing a repo count to create-testbed via setup(), and we could skip setting git users most of the time, but I'd hesitate to add that complexity and testbed variability for the same reasons you mention.

simonharrer:
Hi, 

regarding 1.: Ah, now I get it. The author that will make the actual commit should not be duplicated in the co-authored-by section. Makes sense. Thanks for clarifying. And yes, I agree with using git porcelain for this in combination with golang.

regarding 2.: Sounds good to me. Thanks. Speeding up the tests might be a completely different issue - perhaps even using a check-and-restore logic with the assumption that creating the structure is more time consuming as restoring it from somewhere on the filesystem.

Just a reminder: since 1.1.0 there is `mob done --no-squash` where this change doesn't make any sense as an actual merge commit is created. We need to include your code changes only for the default behavior.

simonharrer:
What’s the status here? Do you need feedback from me somewhere which I missed to give? Just let me know if I can unblock you.

stevegt:
Hi Simon!  Thanks for asking.  The delay is primarily because we're in a minor crunch time ourselves for an upcoming release.  We've also evolved some workflow differences that have us currently using `mob start` but not always `mob next` and rarely `mob done`, and instead manually running git commit, push/pull, and merge.  The differences stem from members' desire for individual attribution, not only in the co-author headers but also in `git blame` results. 

I've barely enough bandwidth right now to analyze how these changes could also be integrated into `mob` as optional behavior; essentially I think it would require not only this PR but a change to `next` to include co-authors as well, a flag to optionally open the `next` commit message in an editor to allow more detail, and a config variable to default to merge instead of squash in `done`.  

I think the tool could do this while preserving current behavior, but that thinking is as far as I've gotten so far.  I'm meanwhile teetering on that dangerous edge of feeling the urge to implement these features quickly in a fork or even in bash scripts, when I think the right thing to do would be to work with you on integrating them.  The latter would obviously take longer to do right, and the upshot is that we're still doing all of these things manually due to this indecision.

In order to break the logjam I'm considering putting all of this together as a coherent group, in one branch, so we can write and use the features ourselves right away from our own fork.  I can feed that branch to you as one PR, either this one or a new one.  If we were to start a new PR, I would call this an `attribution` branch, because that's essentially what this set of changes is about.

Let me know what you think.  There are other wishlist items that we've encountered -- one of them is the idea of a hook or plugin to allow the timer to talk to any external service such that anyone can see current time left and everyone gets the "time's up" message.  Other wishlist items are harder to describe, and would call for more hook architecture because their implementation would be specific to our environment, integration with deploys, discord, etc.  Other than the idea of hooks, I've given near-zero thought to whether a generic tool can satisfy this many org-specific specific requirements, though I'd like to think it could while still remaining simple enough and easy to maintain.

lamalex:
Fwiw I was planning on picking this up. We’ve started experimenting with mob, and we like the flow but co-authors needs implemented before we really start pushing our mobs to use it. I was going to use my 10% time on this, if it needs someone to carry the mantle

simonharrer:
Thanks for your detailed response @stevegt and also @lamalex for offering in helping out here. Let's try to get this sorted:

Everything regarding timer and hooks, I would recommend discussing this somewhere else. We should focus on attribution first, because this seems to be the most pressing issue right now that prevents teams to use the tool.

I see two approaches that are orthogonal:
- Get smaller features from this large thing into the main branch quickly. Reason: I fear long running PRs. We could already add the improved create-testbed for example.
- Discuss the concept we want to implement. I started https://github.com/remotemobprogramming/mob/wiki/Attribution to steer this discussion towards some kind of minimal "spec". Feel free to contribute there.

simonharrer:
@stevegt with `MOB_DONE_SQUASH` you can already disable the squashing on mob done with an environment variable.

lamalex:
The wiki doesn’t seem to have open write access (probably a good thing). Maybe a “discussion” would be an appropriate place to develop a spec in the open?

simonharrer:
let's try that instead https://github.com/remotemobprogramming/mob/discussions/139

stevegt:
Split code from #144 to implement the automatic coauthors portion here -- tests pass and coverage looks decent, but not yet exercised in production -- @simonharrer, @dmcquay, or @lamalex might want to do so; due to local chaos we may not have a chance to exercise this ourselves before the 1.4.0 release.

Here's the commit message:
```
    implement automatic coauthors using selected code from #144
    
    - implements automatic coauthor append, leaves manual
      additions for #144 to implement
    - inspired by @dmcquay bash implementation and closes #81
    - parses SQUASH_MSG using code from @lamalex as opposed to
      the log parsing method described in #81 and at the top
      of #121 -- leaving this as-is to make merging or rebase
      of #144 easier, but if we ever do want to switch to log
      parsing, that change would likely be transparent to users
    
    Co-authored-by: Alex Launi <dev@launi.me>
    Co-authored-by: Dustin McQuay dmcquay@gmail.com
```

stevegt:
Hold up y'all -- while touch-testing I just found a bug in collectCoauthorsFromWipCommits() that is causing co-authors to be missed.  The code currently assumes that the first author matching the regex is the committer, which isn't necessarily true.  I'll want to first write a test case to catch this, and then do something different with that function; either ask `git config` who the committer is, or switch to @dmcquay's method of fetching the coauthors from the git log instead of parsing SQUASH_MSG.  I need to get some other things done right now, but will see if I can get to all of that today.

stevegt:
Okay, better now -- fixed that bug, tests pass including that new test that would have caught the bug, coverage still decent.  Should be ready for 1.4.0 assuming nobody finds anything else this week.

To help verify the new test case is doing enough, I also threw together an extended stress test bash script that generates lots of coauthor commits in a random order -- long-running, needs human intervention, and is possibly destructive, so I'm not proposing it for inclusion in the repo unless we discover more order-dependent coauthor bugs: https://gist.github.com/stevegt/b4562600e047437f54b99ccad43d87d4

mob.go
```diff
@@ -645,6 +645,12 @@ func getCachedChanges() string {
 	return strings.TrimSpace(silentgit("diff", "--cached", "--stat"))
 }
 
+func makeWipCommit() {
+	commitMsg := configuration.WipCommitMessage
+	git("add", "--all")
+	git("commit", "--allow-empty", "--message", commitMsg, "--no-verify")
```
simonharrer, line 651: Should not allow empty commits at the moment. Only necessary for s special case when having the with argument. Can you remove that, please?

stevegt: Yep, missed that -- done!

PR #124, opened 2021-02-07
Removes wipBranchQualifierSet 

gregorriegler:
thus removes a feature where one could start on the mob-session branch despite having another qualified branch on the remote using `mob start ""`

Hi Simon,

I was gonna start and extract the configuration as a parameter in the first command. So I thought I might as well remove `wipBranchQualifierSet` which you said was deprecated and you wanted to get rid of it.
But now that I tried to remove it, I learned what `wipBranchQualifierSet` was really about. The removal kills a feature. So I am not at all confident in that change.

Please have a look. Maybe you do have a better idea.
Could as well keep `wipBranchQualifierSet`, or just remove the feature.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -74,7 +74,6 @@ func getDefaultConfiguration() Configuration {
 		MobStartIncludeUncommittedChanges: false,
 		Debug:                             false,
 		WipBranchQualifier:                "",
-		WipBranchQualifierSet:             false,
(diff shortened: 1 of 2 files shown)


simonharrer:
I see. I have missed this feature. We should keep that feature. Maybe we should document it better so that we don't forget this again?

PR #131, opened 2021-02-20
introduces configuration as parameter to start and subsequent functions

gregorriegler:
improves determineBranches again, but keeping the dualism
testing removeWipQualifier at a better abstraction

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -226,7 +226,7 @@ func execute(command string, parameter []string) {
 
 	switch command {
 	case "s", "start":
-		start()
+		start(configuration)
 		if !isMobProgramming() {
 			return
 		}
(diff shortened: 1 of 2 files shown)


mob.go
```diff
@@ -310,27 +336,16 @@ func removeWipPrefix(branch string) string { //TODO improve, add tests
...
-			return branch[:strings.LastIndex(branch, suffix)]
-		} else {
-			return branch
-		}
-	} else if hasSuffix(branch) { // WipBranchQualifier not configured, but WipBranchQualifierSeparator found
-		return branch[:strings.LastIndex(branch, configuration.WipBranchQualifierSeparator)]
-	} else {
-		return branch
-	}
+func wipBranchQualifierSuffix(configuration Configuration) string {
```
simonharrer, line 347: Such functions could become methods of the configuration struct.

gregorriegler: made it so

PR #132, opened 2021-02-22
next and subsequent functions take configuration as a parameter

gregorriegler:
introduces new method configuration.hasCustomCommitMessage


simonharrer:
Strange that no tests are run in the PR.

gregorriegler:
I could continue on a branch on this repo instead if it makes it easier to verify the build

simonharrer:
would be better

Issue #134, opened 2021-02-26
tests fail with version 1.3


draptik:
While trying to package mob v1.3 from source I encounter the following error:

```sh
go: cannot find main module; see 'go help modules'
==> ERROR: A failure occurred in check().
    Aborting...
```

Building from source is ok.
But running tests currently fails.

```sh
# ....
Creating test assets...

Root of temporary test assets: /tmp/mob
Path to central repository ('remote'): /tmp/mob/remote
Path to clone of central repository ('local'): /tmp/mob/local

Creating test repositories...
Initialized empty Git repository in /tmp/mob/remote/
Cloning into '.'...
warning: You appear to have cloned an empty repository.
[main (root-commit) 4281c7e] initial import
 1 file changed, 1 insertion(+)
 create mode 100644 test.txt
To file:///tmp/mob/remote
 * [new branch]      main -> main
Branch 'main' set up to track remote branch 'main' from 'origin'.

Creating a second mob participant's repository...
Path to second clone of central repository ('localotherdir'): /tmp/mob/localother
Cloning into '.'...
Done.
go: cannot find main module; see 'go help modules'
==> ERROR: A failure occurred in check().
    Aborting...
```

Has running the tests changed in this release?

My build steps currently are:

```sh
./create-testbed # <- no error
go test # <- this throws the error message
```

Details:
- https://github.com/draptik/arch-package-builds/blob/ac89004c00297f6c39724a346c2af917e55d4d4f/mobsh/PKGBUILD#L31
- https://github.com/draptik/arch-package-builds/blob/ac89004c00297f6c39724a346c2af917e55d4d4f/mobsh/PKGBUILD#L34

I have also tried compiling the main branch as described in the README. It returns the same error.

Did I miss something obvious?

simonharrer:
make sure you have go 1.15. It fails with go 1.16.

draptik:
Ok, thanks for the feedback. I'll have to put the source package on hold then, because Arch is a rolling release version (I can't easily use an outdated version of go).
I can still package the prebuild binaries though.
Fun fact: v1.2 already worked with go 1.16...

draptik:
For now I excluded the "test" step for v1.3 (I figured if your test suite passes, I don't have to execute the tests for packaging). AUR packages are up-to-date:
https://aur.archlinux.org/packages/?K=mob-programming&SB=p

simonharrer:
[...]

PR #136, opened 2021-02-27
Truncate list of last commits if it is longer than 5 entries

Morl99:
See https://twitter.com/Morl99/status/1365360716610478080 for the discussion about the feature

I am very new to go and would happily take any comments about how this code could be simpler, if possible. 
I decided to write a test for the "happy" case, where the new code actually makes any difference, and decided that for everything else, the existing tests are good enough.

@simonharrer do you think that this should be configurable behaviour? Can you think of a use case, where mob should consistently print the whole list, or more than 5 entries? 

PS: I was unable to execute the commans that are documented in the "How to contribute section", as Go always complained about a missing module. I readded the module file that you created yesterday and it worked like a charm. I am using go 1.16, not sure if this has anything to do with it. Just wanted to let you know, I did not commit go.mod because I do not want to mess with your project setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.4.0
+- The list of commits included in a mob branch is now truncated to a maximum of 5 entries to prevent the need for scrolling up in order to see the latest included changes.
+
 # 1.3.0
 - The default `MOB_COMMIT_MESSAGE` now includes `[ci skip]` and `[skip ci]` so that all the typical CI systems (including Azure DevOps) will skip this commit.
 - Add `--squash` option to `mob done` that corresponds to `--no-squash`.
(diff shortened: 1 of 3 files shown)


Morl99:
Just came across #134 which explains it all, so just ignore my comment. I read go >= 1.15 in the docs and thought I was good to go with 1.16. Might be safer to document the exact Version that is supported.

simonharrer:
can you double check that the mob status command doesn’t add a empty line if there are no commits? Maybe use a test for that. Otherwise it looks fine for me.

simonharrer:
can you add an entry to the changelog.md targeting 1.4.0?

Morl99:
> can you double check that the mob status command doesn’t add a empty line if there are no commits? Maybe use a test for that. Otherwise it looks fine for me.

That would have been the case before as well, right? You mean the call to say with a potentially empty output?  I can take a look at it, once I come back to my computer (done for the day ;))

Morl99:
I added a test that shows, that no empty line is added, and I included a changelog entry.
Does that work for you?

simonharrer:
LGTM Thanks for your contribution!

PR #138, opened 2021-03-03
Show friendlier error message when run outside of git repo

### LINKED ISSUE #114: Add better message when issuing a mob command outside a git repository

lamalex:
Use git rev-parse to detect if mob is run from inside of
a git repo. Inform the user they may not be where they
thought they were.

Message reads: 
> 🤦🏿 It doesn't appear that you're running `mob` inside of a git repo ...
>       Check that you're in the right place.

ezgif com-gif-maker

Closes #114


simonharrer:
Awesome. Curious: how did you create that gif?

Regarding your change: is fine but I would use a more concise error message:

mob expects the current working directory to be a git repository. The current working directory INSERTDIR is not a git repository.

lamalex:
1) I used screen recorder, then used https://cloudconvert.com/mp4-to-gif to convert it to a gif, and finally used https://ezgif.com/speed to speed it up 400% because it was *so slow*

Message updated.
The only thing I'm not happy with is the test since the actual working dir is not changed I didn't assert that the message prints the working directory.

simonharrer:
Thanks for the info. 

Yes, I understand regarding the test. Would be fine by me though. Can you add a changelog.md entry please? Then I can merge this and it will be part of the 1.4.0 release. :-)

PR #142, opened 2021-03-05
Added nix expression to readme.

jeeger:
Added nix expression to readme.


simonharrer:
could you put the nix expression in a separate file and only reference that file from the readme?

simonharrer:
Thanks for your help! :-)

PR #144, opened 2021-03-07
Add coauthors using --with <...>

### LINKED ISSUE #81: Automatically Populate Co-Authored-By lines

lamalex:
We love the workflow `mob` enables but we need co-author attribution, and to support attribution for members of a mob who may not have actually written any code but were present. From discussion in #139, we brainstormed and came up with a workflow we think will be very nice.

Our idea is this:
Add a `--with` option to `mob start` which takes a comma separated list of coauthors and/or shorthand-aliases like, 
```bash
$ mob start --with "Ibram X Kendi <ixk@x.com>, Amanda Huggenkiss <ah@x.com> as hugz, sh"
```

'Ibram X Kendi' would be saved as 'ixk' for future shorthand reference, 'Amanda Huggenkiss' as 'hugz', and sh references an already stored alias.

Aliases are permanently stored in the global .gitconfig in a [mob] section.
When `mob start --with <...>` is run the coauthors are parsed and added to [mob "wip"] subsection.
When `mob next` or `mob done` is run the coauthors listed in the [mob "wip"] subsection are appended as coauthors to the WIP commit.

At `mob done` all authors, and co-authors are parsed, collected, and printed to the squash commit.

Before merging
============
 - [x] Parsing --with
   - [x] Fully qualified author + email
   - [x] Author without alias -> alias generation
   - [x] Solo aliases    
 - [x] Saving aliases to git config
 - [x] Expanding aliases from git config
 - [x] Writing coauthors to WIP commit
 - [x] Collecting authors/coauthors into final commit.
 - [x] Refactor to support utf-8 (I wrote the lexer with no internet between 8pm and 5am and I just string sliced)
 - [x] Add comment to squash commit regarding automatic append of coauthors
 - [x] Clear staged coauthors after `mob reset`
 - [x] Respect `mob start` after `mob start`  
 - [x] Deduplicate co-authors

Fixes #81


lamalex:
@simonharrer there are some pieces of this that could be split and merged separately- for instance @stevegt ’s testbed changes, and the code to avoid relying on “master”. Let me know if you’d like a separate PR with those changes isolated

simonharrer:
Thanks for your initiative.

I'd like to do a review but the creation of new modules makes it hard for me to see the actual changes in the code as so many lines are marked as changed. Can you please revert the extraction of the new modules - we can do this later if this makes sense. I just want to focus on the new feature in the review.

Also, things that are isolated I would love as separate PRs if this is no hassle for you.

I like using git to store infos from mob. Haven't thought about this before.

lamalex:
Oh snap. My bad. I thought that I was going to use Say and some Git in the parser/lexer and I ended up keeping it separate. I'll clean that up thank you!

lamalex:
> I haven't had a chance to look at or try the code yet, but in principle this UI sounds like it would work for us too. I'm glad @lamalex has been able to tackle it.

@stevegt would you pull out the TestStartDoneCoauthors test and rename your PR for just the test bed fixes? I didn't include any tests for the actual squash msg or wip commit because my original branch relied on yours, but I wanted to keep the pieces separate for @simonharrer 's review sanity

simonharrer:
Okay, the picture is becoming clearer.

Hm, I don't see a real value in the `gitconfig()` method. Just using the `git()` method like this `git("config", "--global", fmt.Sprintf("mob.%s", alias), coauthor)` should probably suffice.

Before getting really into the code, I'd would like to know how the feature would be used. Here's a typical workflow with a mob of three. 

```
carola $ mob start
carola $ # does changes
carola $ mob next

maria $ mob start
maria $ # does changes
maria $ mob next

simon $ mob start
simon $ # does changes
simon $ mob next

# second round starts

carola $ mob start
carola $ # does changes
carola $ mob next

maria $ mob start
maria $ # does changes
maria $ mob next

simon $ mob start
simon $ # does changes
simon $ mob done --squash
simon $ git commit -m "some message"
```

How would you in your team use the `--with` option and when in the above scenario? What would be side effects, meaning, which commits would have which co-authors? And you add this info in the code snippet, please?

lamalex:
@simonharrer gitconfig was added because

1. git() doesn't return anything so I couldn't use it to get saved aliases
2. git config exits with 1 if nothing is found, which causes silentgit to exit

and I didn't want to completely refactor the way that those worked.

lamalex:
```
carola $ mob start --with "maria, simon" 
carola $ # does changes
carola $ mob next

# commit looks like:
# Author: Carola <>
# Date:   Thu Mar 11 16:45:22 2021 -0500
#
#    mob next [ci-skip] [ci skip] [skip ci]
#
#    Co-authored-by: Simon <>
#    Co-authored-by: Maria <>
 
maria $ mob start
maria $ # does changes
maria $ mob next
 
simon $ mob start
simon $ # does changes
simon $ mob next
 
# second round starts
# Alex joins the mob
 
carola $ mob start --with "Alex L <dev@launi.me> as alex"
carola $ # does changes
carola $ mob next

# commit looks like:
# Author: Carola <>
# Date:   Thu Mar 11 16:45:22 2021 -0500
#
#    mob next [ci-skip] [ci skip] [skip ci]
#
#    Co-authored-by: Alex L <dev@launi.me>
 
maria $ mob start
maria $ # does changes
maria $ mob next
 
simon $ mob start
simon $ # does changes
simon $ mob done --squash
# commit -m ignores .git/SQUASH_MSG so this doesn't actually append coauthors
# Using -m requires a pre-commit hook
# simon $ git commit -m "some message"
simon $ git commit

> Solve halting problem
>
>    Co-authored-by: Maria <>
>    Co-authored-by: Simon <>
>    Co-authored-by: Alex L <dev@launi.me>
```

simonharrer:
Okay, now I get the full picture.

In case of

```
carola $ mob start --with "maria, simon" 
carola $ # no changes made
carola $ mob next
```

there won't be a commit and the `mob.staged` section would still exist locally. Not sure if this is the expected behavior. Perhaps a `mob next` should always clear the `mob.staged` section but might issue a warning that the co-authors were not added to any commit.

simonharrer:
What about adding a comment in the commit message saying something like this:

# mob automatically added all co authors of the WIP commits
# add missing co authors manually

lamalex:
> Okay, now I get the full picture.
> 
> In case of
> 
> ```
> carola $ mob start --with "maria, simon" 
> carola $ # no changes made
> carola $ mob next
> ```
> 
> there won't be a commit and the `mob.staged` section would still exist locally. Not sure if this is the expected behavior. Perhaps a `mob next` should always clear the `mob.staged` section but might issue a warning that the co-authors were not added to any commit.

I would expect mob to make a commit with `—allow-empty` and the coauthors. This isn’t currently handled, but it should be. Thanks; I’ll push an update in a bit.

lamalex:
> What about adding a comment in the commit message saying something like this:
> 
> # mob automatically added all co authors of the WIP commits
> # add missing co authors manually

Do you mean in the final squash commit?

simonharrer:
> Do you mean in the final squash commit?

yes, in the final squash commit.

simonharrer:
> > Okay, now I get the full picture.
> > In case of
> > ```
> > carola $ mob start --with "maria, simon" 
> > carola $ # no changes made
> > carola $ mob next
> > ```
> > 
> > 
> > there won't be a commit and the `mob.staged` section would still exist locally. Not sure if this is the expected behavior. Perhaps a `mob next` should always clear the `mob.staged` section but might issue a warning that the co-authors were not added to any commit.
> 
> I would expect mob to make a commit with `—allow-empty` and the coauthors. This isn’t currently handled, but it should be. Thanks; I’ll push an update in a bit.

Not a great fan of those empty commits, as they have their very own issues - I have tried using them but there was strange behavior like when two people one after another would do an empty commit didn't really work well (but perhaps it was just my programming...). But probably helpful just in case you have `mob.staged` co-authors and use `mob next`. Otherwise, I wouldn't use it, or have quite some tests. 

Also, a `mob start` without the `--with` attribute should probably reset the `mob.staged` attribute, shouldn't it?

simonharrer:
Also, `mob reset` should reset the `mob.staged` co-authors as well.

simonharrer:
start after start is valid.

Another thought: mob.staged could be timestamped so it automatically expires on the next day? Or is this overengineered? Not sure here.

stevegt:
> @stevegt would you pull out the TestStartDoneCoauthors test and rename your PR for just the test bed fixes? I didn't include any tests for the actual squash msg or wip commit because my original branch relied on yours, but I wanted to keep the pieces separate for @simonharrer 's review sanity

Sorry for the delay:  All is chaos here at the moment.  You want me to just refactor my existing `t7a:append-coauthors` branch so  it flows into #121, or can I create a new branch and PR that just includes the new testbed, and you rebase onto that?

lamalex:
> > @stevegt would you pull out the TestStartDoneCoauthors test and rename your PR for just the test bed fixes? I didn't include any tests for the actual squash msg or wip commit because my original branch relied on yours, but I wanted to keep the pieces separate for @simonharrer 's review sanity
> 
> Sorry for the delay: All is chaos here at the moment. @lamalex Do you want me to just refactor my existing `t7a:append-coauthors` branch so it flows into #121, or can I create a new branch and PR that just includes the new testbed, and you rebase onto that?

@stevegt Whichever is easier for you I can work with. The latter might be easier for me to combine my changes with though.

lamalex:
A coworker and I paired on this while we dog-fooded it today.
We updated `mob start` to set an option on `configuration` which defaults to an empty map and then populates with the contents of `--with` so start will always overwrite staged with whatever is given by `--with`.

We went back and forth on empty commits and we didn't have any problems. We passed off `mob start --with "x <>"; mob next` a handful of times. @simonharrer what sort of weird behavior did you observe prior? To be safe we set it to only make an empty commit if there are staged coauthors. Otherwise it maintains current behavior of no commit.

simonharrer:
I would like to user to understand this feature. For that we should print any change to mob.staged on the console (setting and resetting) and describe the setting and resetting relationship in the help section.

lamalex:
Ok, some output has been added:

### There are *no* coauthors staged in ~/.gitconfig
```
$ mob start
 > (normal co-authorless output)
```

### There are *no* coauthors staged in ~/.gitconfig
```
$ mob start --with "Spike Lee <5bloods@40acres.movie>"
 > (normal co-authorless output)
 > Spike Lee <5bloods@40acres.movie> was saved to ~/.gitconfig
    Next time you can use `mob start --with "sl"`
 > Spike Lee <5bloods@40acres.movie> has been staged as coauthors in ~/.gitconfig
    They will appear as co-authors on your next WIP commit,
    and they will appear as co-authors after `mob done`.
```

### There are *no* coauthors staged in ~/.gitconfig
```
$ mob next
 > (normal co-authorless output)
```

### There are *no* coauthors staged in ~/.gitconfig
```
$ mob reset
 > (normal co-authorless output)
```

### There are *no* coauthors staged in ~/.gitconfig
```
$ mob done
 > (normal co-authorless output)
```

### There *are* coauthors staged in ~/.gitconfig
```
$ mob start
 > (normal co-authorless output)
 > Cleared previously staged co-authors from ~/.gitconfig
```

### There *are* coauthors staged in ~/.gitconfig
```
$ mob start --with "Spike Lee <5bloods@40acres.movie>"
 > (normal co-authorless output)
 > Cleared any previously staged co-authors from ~/.gitconfig
 > Spike Lee <5bloods@40acres.movie> has been staged as coauthors in ~/.gitconfig
    They will appear as co-authors on your next WIP commit,
    and they will appear as co-authors after `mob done`.
```

### There *are* coauthors staged in ~/.gitconfig
```
$ mob next
 > (normal co-authorless output)
 > Co-authored-by: Spike Lee <5bloods@40acres.movie>
```
**Note**: I think in your comment you wanted `mob` to print that it was clearing staged coauthors on `next` as well, but when I tried it like this I actually found it confusing. Saying nothing on `mob next` felt more natural. Clearing authors on `mob next` is an implementation detail and printing it felt leaky.

### There *are* coauthors staged in ~/.gitconfig
```
$ mob reset
 > (normal co-authorless output)
 > Cleared any previously staged co-authors from ~/.gitconfig
```

### There *are* coauthors staged in ~/.gitconfig
```
$ mob done
 > (normal co-authorless output)
 > Cleared previously staged co-authors from ~/.gitconfig
```

simonharrer:
Great output! :-)

How important is the alias feature for you? Because lexer and parser require quite some code and an ever growing lookup table is stored in git config under the key „mob“. My idea would be to only let the user pass in a comma separated list of co authors and we just set this list as a string to mob.staged in the git config and reuse this string for mob next. A lot less code and much simpler. If one would mob in a specific group they could use a bash alias or script for that mob start call. Perhaps even the took git mob could manage those aliases and we could integrate those tools? What do you think?

Regardless, we need some acceptance tests similar to the existing ones.

lamalex:
The aliasing is actually pretty important to me. And while it is much less code not to, the parser and lexer are a pretty basic standard lexer/parser which and have their own set of tests, and is well tested. I don't think it adds an unreasonable amount of complexity to the tool itself, and the complexity that it *does* add is contained inside of functions meant to deal with --with. Not leaked throughout.

As for "mob"  section in .gitconfig ever growing that is a concern, but do users mob program with hundreds of different developers? I don't think in reality that section grows much better than 50 - which is big - but is there a real cost?

I agree it needs acceptance tests -- I was hoping @stevegt would clean his branch up so that the testbed changes could be merged separately. I'l try and take some time tonight to get his testbed changes ready for merging separately.

stevegt:
@lamalex I should be able to get to it within a few hours (Sunday evening California time) if you don't get there first.

stevegt:
@lamalex Actually, a time window opened up for me just now.  Diving in...

stevegt:
@lamalex take a look at #146 -- I'll be around for the next few hours.  Tomorrow back to the same chaos, but this should give you enough of a base to build on.

simonharrer:
@lamalex okay fine by me. Was just wondering if it could be simplified. 

I have a few minor things. Will try to either put it in here as a code comment or change it myself afterwards.

I think we can ship it with 1.4.0 in the next release. Please add a changelog entry describing the new feature.

coauthors/lexerparser.go
```diff
@@ -0,0 +1,254 @@
...
+	p := newParser(l)
+
+	return p.parseCoauthors()
+}
+
+// CoauthorsMap is a mapping from --with supplied
+// alias/author name. Alias will always be
+// a non-empty string. Author may be an empty
+// string for solo-aliases
+type CoauthorsMap = map[Alias]Author
```
simonharrer, line 23: Can we move the data structure to the main go file? So that the lexer/parser only contains the parsing? Not sure if go can handle this...

mob.go
```diff
@@ -645,6 +680,21 @@ func getCachedChanges() string {
 	return strings.TrimSpace(silentgit("diff", "--cached", "--stat"))
 }
 
+func makeWipCommit() {
+	git("add", "--all")
+	wipCommitMsg := configuration.WipCommitMessage
+	stagedCoauthors := gitStagedCoauthors()
```
simonharrer, line 686: can you extract a method that gets the configured wip commit message and adds the coauthors? Mabye name it addCoAuthors?

mob.go
```diff
@@ -990,3 +1063,191 @@ func sayEmptyLine() {
...
+			if topLevelAuthor == "" || author == topLevelAuthor {
+				topLevelAuthor = author
+				continue
+			}
+			coauthorsHashSet[author] = true
+		}
+	}
+
+	if len(coauthorsHashSet) > 0 {
+		coauthors := make([]string, 0, len(coauthorsHashSet))
```
simonharrer, line 1218: would love a method that converts co authors to a git commit message string. So we can separate the logic from the file append.

coauthors/lexerparser.go
lamalex: @simonharrer We can't move the CoauthorsMap into `mob.go` because it will create a cyclic dependency, but we could import coauthors like

```go
import (
    ...
    
    . "mob.sh/coauthors"
)
```
and import coauthors' members directly into mob without the namespacing

lamalex:
> I think we can ship it with 1.4.0 in the next release. Please add a changelog entry describing the new feature.

Added!

simonharrer:
I think we are almost done. My plan is to release 1.4.0 next week. Anything you still miss @lamalex ? How about you @stevegt ?

@gregorriegler would you like to have a go at a review of this feature? More eyes see more. :-)

lamalex:
It needs some more acceptance tests, but the feature is complete

stevegt (changes requested):
Code review only -- code not executed yet.  In summary, I'd deprecate the coauthors and testutils packages and use RFC-2606-compliant email addresses in examples.

stevegt:
Yeah, we're missing test cases, I think.  @lamalex are you working on those?  (@gregorriegler dinged me for test coverage on my first PR here -- caused my entire team to switch to more consistent TDD so we're better for it.)

I think we never got some language into README.md for guidance, but you can find @simonharrer's coverage intent at https://github.com/remotemobprogramming/mob/pull/107#issuecomment-759297513.

coauthors/lexerparser.go
```diff
@@ -0,0 +1,254 @@
...
+func (p *parser) expectPeek(t TokenType) bool {
+	if p.peekTokenIs(t) {
+		p.nextToken()
+		return true
+	}
+
+	return false
+}
+
+func generateInitialation(author Author) Alias {
```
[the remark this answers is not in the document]
lamalex: Good catch: `initialism` was the term I was reaching for -- but generateInitials is much clearer.

stevegt:
Looks pretty clean.

I do get test failures on the first few runs, as the new config machinery works its way through initial setup.  You'll probably want to rm or mv `/tmp/mob` out of the way manually and see what that looks like.  (The create-testbed code uses `/tmp/mob` instead of a unique tempdir on each run -- we don't `rm -rf /tmp/mob` automatically, because that's a bit risky, particularly since a bunch of people would `git clone https://github.com/remotemobprogramming/mob /tmp/mob` for a test drive before installing.  I already have an `XXX` mark in create-testbed to remind us to fix this someday, but that should be handled in a separate issue/PR; it's going to need mob_test.go to call os.TempDir() and pass the path as an argument to create-testbed.)

A test case for `--with` would improve coverage, but I haven't stared at the code long enough to figure out how critical that is.

lamalex:
Added test for parsing --with.
I am a little concerned about some of the tests for dealing with gitconfig. We're operating on the global git config, and performing potentially destructive operations. For instance if you were mob programming on mob with staged coauthors and you ran the test you would blow away your staged coauthors.

Not sure what the solution is. Cache and write back any staged coauthors? Enough of an edge case (and solely a developer issue) that it's ok?

stevegt:
Oh! I hadn't noticed we're hitting the global config. I'm on my feet and
can't look at the code right now, but in general I would strongly suggest
we only use local. In my case I wear a lot of hats and need to be able to
keep different projects firewalled from each other. That would take huge
precedence over the inconvenience of needing to configure mobbers in each
repo; in most cases they aren't going to be the same people in the first
place.

stevegt:
Restating my previous message more concisely -- infosec requirements would
call for erring on the side of local config.

simonharrer:
I think the alias lookup table makes sense to be global. We do have a lot of git repos and it would be annoying to add the authors for every git repo we would work together. 

The staged co authors, however, should probably be local. As the with option is meant for mob next in the same repo.

Thanks for catching that to both of you! :-)

lamalex:
>  In my case I wear a lot of hats and need to be able to keep different projects firewalled from each other. That would take huge precedence over the inconvenience of needing to configure mobbers in each repo; in most cases they aren't going to be the same people in the first place.

Any additional data that is this true? I mob with some 2-4 person combination of my 10 person team almost exclusively. Which combination changes, but the set of potential mob partners is small, and is spread across a variety of repos such that having to re-alias would actually be a burden. Granted it would be once per project, that but that could still be over a dozen times for us.

I'll let @simonharrer give an opinion before I change anything. Initially @simonharrer liked using git config to store the data (as do I). I would be open to adding an envvar to tell mob that it should not use global, and to use a specific file. something like `MOB_COAUTHOR_CFG=.git/config` or `MOB_COAUTHOR_CFG=.mob-coauthors` but when unset just uses global.

lamalex:
my initial thinking was that the [mob.staged] section is highly transient. it only lasts between mob start and mob next, but I think that the right answer is to keep it local. it **is** possible to be mobbing on multiple related projects at the same time and you wouldn't want mob next in one to wipe staged coauthors before mob next in another.

gregorriegler:
Sorry, i didn't noticed i was pinged here. Now trying to read through this.

stevegt:
> I would be open to adding an envvar to tell mob that it should not use global, and to use a specific file. something like `MOB_COAUTHOR_CFG=.git/config` or `MOB_COAUTHOR_CFG=.mob-coauthors` but when unset just uses global.

That's what I was about to suggest, but I would default it to local, not global.  Defaulting to global screams badness to me -- it's going to result in internal members of private or even classified projects accidentally getting published on github as coauthors in open-source repos that they have had nothing to do with.

> > In my case I wear a lot of hats and need to be able to keep different projects firewalled from each other. 
> 
> Any additional data that is this true? I mob with some 2-4 person combination of my 10 person team almost exclusively. 

I context-switch between open source projects, multiple internal project-specific repos, and an internal monorepo.  Team members may or may not overlap, and may be members of other organizations.

I'd suggest that, if you have multiple repos but one team, then you may be needing to maintain excessive release machinery and documentation, may be encountering version mismatch errors as repos release at different times, and may be a candidate for a monorepo instead.  If true, then a global git config is a workaround for a deeper problem.  Defaulting to a global config would be the wrong way of approaching that -- let the default be local, so it's compatible with common practice in solving that problem, and use `MOB_COAUTHOR_CFG` to allow for the workaround in those cases when people can't move to a monorepo for political or organizational reasons.

lamalex:
I think we all agree the actual staging area should be project local, but I don't see how your scenario has a problem with the global alias lookup.

gregorriegler:
Nice feature. Sad that git doesn't support for n>1 author commits natively.

Not sure if the global aliases would be a problem.
I use a different mail (and thus Git Author Handle) for my job than I use for private and open source work, and I hate it when I accidentally mix those.
I know people who work for different clients and use different emails for each, so I see where @stevegt is coming from.

Personally, I would choose to start it as a minimum viable feature (without the global aliases) and wait for feedback in this regard. Maybe it's not a problem at all.
I am already used to create .gitignore files for each repo and don't mind managing coauthors for each repo individually either, as I'd have to do it very rarely.

We need to explain this feature very well in the documentation. I think it's not obvious to users how it behaves.

I like that we're starting to create new files/modules, `mob.sh` and `mob_test.sh` became a bit long.

Are we sure we have covered every possible use case with a test?

lamalex:
So I think maybe some clarification needs made on how the feature works would be useful.
For most users it's transparent; they have no need of using `--with`.

For example:
If `n` users are mobbing and each makes a WIP commit `n-1` people will appear with `Co-authored-by:` lines in the squashed message, and 1 (the `mob done` user) will be the committer. 

`--with` covers something of an edge case (that my team actually faces fairly often). There's a case where not all members of the mob actually write any code. For example we mob/pair exclusively. Often we are in a mob of 3 working on something extremely simple that takes say 15 minutes of keyboard time. In this case only 1 member of the mob will interface with the code. In order to **ensure** co-author attribution --with comes into use. Sometimes our tech lead is "in the mob" but she is just providing some high level guidance. We want to include her on commit attribution although she's not really part of the mob, etc.

gregorriegler:
hm. so adding a non-commiting mob-member is now coupled (only possible) with `mob start`. do we want that? is that the best option?

lamalex:
> hm. so adding a non-commiting mob-member is now coupled (only possible) with `mob start`. do we want that? is that the best option? `start` can be entered twice, but how does it feel to enter `start` a second time for a late `with`?

Could you spell out the use case you're imagining a little more? I don't think I understand

stevegt:
@lamalex You'd want to avoid people saying "You've modified my global git config without my consent or warning."  If people want their ~/.gitconfig modified, that should be their choice, as an opt-in.  (It was actually pretty disconcerting to me just now to check my own ~/.gitconfig on my laptop and find that you'd modified it, so there's that.)

In our case, if the default is not local, then that adds workload as well -- we'd have to make sure that everyone sets and maintains the MOB_COAUTHOR_CFG variable in all of their shells on all of their machines in order to avoid accidental info leakage.  

If I understand the UI correctly, a mis-remembered set of initials can land the wrong person's email address in a git commit message, right?  I'm not so worried about that for open source projects, but the potential for leakage from internal addresses is not good.

stevegt:
> `--with` covers something of an edge case (that my team actually faces fairly often). There's a case where not all members of the mob actually write any code. For example we mob/pair exclusively. Often we are in a mob of 3 working on something extremely simple that takes say 15 minutes of keyboard time. In this case only 1 member of the mob will interface with the code. In order to **ensure** co-author attribution --with comes into use. Sometimes our tech lead is "in the mob" but she is just providing some high level guidance. We want to include her on commit attribution although she's not really part of the mob, etc.

I agree with and support this feature 100% -- we have the same need for the same reasons.  I also can see that having a single set of initials to remember, in a global ~/.gitconfig, is in some ways safer than using only local configs, where the same initials might map to different people in different repos.  There may be a third alternative we haven't seen yet, but in the meantime I think altering the global config by default makes both testing and safe usage harder.

lamalex:
There is the option of calling `git config -f ~/.mob-coauthors ...` or something like it.
This maintains a global alias list, but avoid modifying a user's global git config. @stevegt do you think that this would meet your needs while maintaining the flexibility of a global alias list?

stevegt:
Oh -- here's another reason to not use the global .gitconfig:  The test cases will wipe real user data.  Try this:

- add this to your ~/.gitconfig: 

 ```
[mob]
    alice = Alice Hargreaves <alice@example.org>
```

- now run `go test`.  Your ~/.gitconfig will now contain this:

```
[mob]
    alice = alice <alice@example.com>
    bob = bob <bob@example.com>
```

lamalex:
> Oh -- here's another reason to not use the global .gitconfig: The test cases will wipe real user data. Try this:
> 
> * add this to your ~/.gitconfig:
> 
> ```
> [mob]
>    alice = Alice Hargreaves <alice@example.org>
> ```
> 
> * now run `go test`.  Your ~/.gitconfig will now contain this:
> 
> ```
> [mob]
>     alice = alice <alice@example.com>
>     bob = bob <bob@example.com>
> ```

I know! That's how we ended up here! See https://github.com/remotemobprogramming/mob/pull/144#issuecomment-803022034

gregorriegler:
> > hm. so adding a non-commiting mob-member is now coupled (only possible) with `mob start`. do we want that? is that the best option? `start` can be entered twice, but how does it feel to enter `start` a second time for a late `with`?
> 
> Could you spell out the use case you're imagining a little more? I don't think I understand

the case is: bob has already `mob start` and alex joins late. bob can now run start again like so: `mob start --with alex` right? how does `mob start` for adding people late feel? compared to 

`mob start`
`mob join alex`

stevegt:
> I know! That's how we ended up here! See [#144 (comment)](https://github.com/remotemobprogramming/mob/pull/144#issuecomment-803022034)

Yep -- this has all been my response to the concern you raised in that comment.  I'm glad you mentioned it, and I apologize for taking this long to fully understand what you were concerned about.  I think my overtaxed brain started from an invalid assumption that when you said "global git config" you were already setting GIT_CONFIG to something out of the way in the test cases so they wouldn't modify production data, and not actually modifying my own live ~/.gitconfig when I ran `go test`.  Me gaining a full understanding of what's really going on has been a slow drip of "oh wait, it really is...".

I see now in the git-config man page that even if you were setting GIT_CONFIG to something innocuous, the --global flag overrides that anyway.  I think a "smell test" for fixing all this would be that --global shouldn't appear anywhere in the code or test cases -- it's too risky and can't be overridden even if the user needs it to be.

> I am a little concerned about some of the tests for dealing with gitconfig. We're operating on the global git config, and performing potentially destructive operations. For instance if you were mob programming on mob with staged coauthors and you ran the test you would blow away your staged coauthors.
>
> Not sure what the solution is. Cache and write back any staged coauthors? Enough of an edge case (and solely a developer issue) that it's ok?

The solution is to not default to using the global config -- too much can go wrong, including things we cannot imagine ourselves.  If you want to have a `MOB_COAUTHOR_CFG` env variable that people can set if they do want to use the global config, that's a good idea.

But testing against the users' personal global config can never be done safely in a production machine without making unsafe assumptions about how different people manage their world.  A test case for `MOB_COAUTHOR_CFG` would have to aim at another location for the config file, never ~/.gitconfig.  To test `MOB_COAUTHOR_CFG`, I'd set it to `/tmp/mob/notgit/.gitconfig` for now, with a note in #149 to make sure that gets changed to the new tempdir created after create-testbed is fixed.

> I actually tried to avoid using aliases that might be common (Alice). the others are mob-, but I seem to have missed this one.

It's not possible to safely guess what aliases will always be unused in production; this is an extension of the halting problem.

> Damn you beer!

I could always say I told you so.  ;-)

stevegt:
> There is the option of calling `git config -f ~/.mob-coauthors ...` or something like it.
> This maintains a global alias list, but avoid modifying a user's global git config. @stevegt do you think that this would meet your needs while maintaining the flexibility of a global alias list?

That could work, but a more general solution would be to go one step further and establish a ~/.mob/ directory right now (and a MOB_DIR var) so we can put other persistent blobs in there, things that don't fit in a config file format.  That would let you do whatever you want with format, and free you from the constraints of going through the git config machinery in the first place.  You could, for instance, just store the CoauthorsMap as JSON in ~/.mob/coauthors, along with the aliases table in either the same or a different file; that would give us a lot more flexibility and remove some of the conflicting requirements we're running into here.

stevegt:
@lamalex I just edited a typo in my previous message (about JSON file content), so you'll want to check https://github.com/remotemobprogramming/mob/pull/144#issuecomment-803429152 if you're following this in email.  

Otherwise, looking through the code again, I'm starting to suspect that there might be a lot less code if you were to go the JSON route. For instance, most of the 53-line writeCoauthorsToGitConfig() function collapses down to a json.Marshal(), I think.

simonharrer:
I‘d like to release 1.4.0 next week. The storing of info seems to be an issue which needs more discussion. I’d like to split this PR. The automatic adding of all authors of wip commits doesn’t require any storage in a git config file or somewhere else. This I would like to ship with 1.4.0. Everything else I would like to have in a new PR where we can discuss this further. Okay for you?

stevegt:
A quick and dangerous run through the code tells me that leaving the appendCoauthorsToSquashMsg() call in mob.go, the TestStartDoneCoAuthors() function in coauthors_test.go, and then adding code from coauthors.go until it builds and passes might be a way to get the split without too much trouble.  Does that sound about right @lamalex?

stevegt:
I'll have a go at updating #121 to do that half of the split -- I expect some distractions today but we'll see if I can get this done by this evening PDT.

lamalex:
Yeah that sounds about right. Most of coauthors.go deals with saving and loading aliases so really shouldn’t need much of it iirc

stevegt:
Well, that wasn't too horrible -- done already.  @lamalex I think a rebase or merge of your `--with` code on top of #121 might actually be pretty clean -- I tried to minimize potential merge conflicts but did have to move the Author declaration to coauthors.go, tweak makeWipCommit(), stuff like that.  

@simonharrer I think #121 might be usable for 1.4.0.

stevegt:
@simonharrer Hang on -- found a bug and will track progress in #121.

Morl99:
Wow, what a lively discussion. I just wanted to add my opinion about modifying the global git config: If that happened implicitly, it would really confuse me, as I work with multiple global git files, so I would vote for a local alias. 

Or would it make sense to manage aliases explicitly with a separate command? It might make the usage more intuitive. Then --local and --global could control the storage behavior.

lamalex:
So I think consensus is coalescing around: don't use global .gitconfig, but global aliases are preferred so use an external file (such as .mob/aliases.json) and then we can remove quite a bit of code related to (de)serialization of aliases.

lamalex:
Been very busy. Will. have some time on Friday to work on this! Not forgotten/abandoned!

simonharrer:
I’m thinking we are more and more tending to reimplement the tool „git mob“ as this tool stores the alias to author map in a json file and also provides a way to set the current co authors. I am more and more thinking of simply providing a way so that both tools work well together.

lamalex:
I think you're right. I think just doing parsing of committers from coauthors covers a large surface area of use cases. If that's merged I think we can actually close this pr

PR #145, opened 2021-03-10
Use git-config default branch name rather than hard coding master

lamalex:
Many users have init.defaultBranch set to main, or something other than master.
Tests fails for those users. This makes the test suite respect non-master default branches.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -312,8 +312,9 @@ func execute(command string, parameter []string) {
 }
 
 func determineBranches(currentBranch string, localBranches []string, configuration Configuration) (baseBranch string, wipBranch string) {
-	if currentBranch == "mob-session" || (currentBranch == "master" && !configuration.customWipBranchQualifierConfigured()) {
-		baseBranch = "master"
+	defaultBranch := gitDefaultBranch()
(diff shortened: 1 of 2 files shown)


simonharrer:
The normal pattern is `branchname` as the base branch and `mob/branchname` is the wip branch. The `master` `mob-session` is a special relationship. It's kept for backward compatibility. If you use another default branch like `main`, the wip branch should be `mob/main` and not mob-session.

That being said, I think the PR went into the wrong direction.

However, it uncovered an important assumption that failed on your machine. The tests fail if you have different git settings `defaultBranch`. The problem, I think, lies in the testbed shellscript. At the moment, the testbed and the tests assume the defaultBranch is `master`. The testbed should take this into account when creating the git repositories. I haven't looked but perhaps one can specify the default branch when creating a new git repository? Here, we could explicitly say `master` as all the tests depend on this assumption. We can change it later to main when we want to invest the time to switch.

Does this makes sense?

lamalex:
I .. think so? but actually the test are now passing fine so I may have broken something totally irrelevant and then gone out of my to fix it. I'm going to close this PR ..sorry!

PR #146, opened 2021-03-15
refactor create-testbed to create more repos

stevegt:
- in support of coauthors efforts (#144, #121, #81) per request from @lamalex 
- also creates the /tmp/mob/notgit subdir
- includes a simple test case for the testbed itself
- includes the code allowing run() to return a string, and a test case for same


simonharrer:
Thanks for that! :-)

PR #148, opened 2021-03-16
Refactor to configuration param

gregorriegler:


simonharrer:
Is the configuration now no longer a global variable?

gregorriegler:
> Is the configuration now no longer a global variable?

still is, but the overall goal would be that it isn't.

PR #150, opened 2021-03-21
Remove global variable configuration (still in tests though)

simonharrer:


mob.go
```diff
@@ -224,7 +220,7 @@ func config(c Configuration) {
 	say("MOB_TIMER" + "=" + c.MobTimer)
 }
 
-func parseArgs(args []string) (command string, parameters []string) {
+func parseArgs(args []string, configuration Configuration) (command string, parameters []string, configuration2 Configuration) {
```
gregorriegler, line 223: not sure if we need the configuration2 here. but I can fix that later

gregorriegler (approved):
awesome! thank you for finishing my job :D 
when I find time again, I would like to try and put the configuration-related code into its own file. what do you think?

mob.go
simonharrer: I don’t know enough Go to really know a better way...

Issue #156, opened 2021-03-29
mob does not include unpushed commits in the mob branch


TrentonAdams:
All of these except `blah` were in my .gitignore.
```
$ git status
A  .idea/limited-wip.xml
A  blah
A  good.sh
A  junit-migrate.sh
A  rmi/apas-dataset-synchronizer.log
A  rmi/apas-dataset-synchronizer.log.debug
A  selenium/acceptance/.env.old
```

All I did was...

```bash
mob start
echo hello > blah
mob done
```

And mob.sh decided it should add files that are explicitly ignored.  That seems non-functional.

TrentonAdams:
Oh, it's worse than that.  I started a mob session without first pushing my changes up, so it appears it did not notice that, and decided to create the branch off the origin branch, which did not include my .gitignore changes. 

I'm going to take a look at the mob.sh code.

simonharrer:
Can you rename this issue, please. It’s somewhat misleading for others because it’s not an issue with git and gitignore files but with unpushed local commits. :-)

PR #157, opened 2021-03-29
issue #156 - prevent mob session with outstanding local commits

TrentonAdams:
Let me know if this makes sense as a first start to fixing this issue.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -541,7 +541,7 @@ func start(configuration Configuration) {
 		git("stash", "pop", stash)
 	}
 
-	if hasLocalCommits(currentBaseBranch, configuration) {
+	if hasUnpushedCommits(currentBaseBranch, configuration) {
 		sayError("cannot start; unpushed changes on current branch must be pushed upstream")
 		return
 	}
@@ -749,6 +749,17 @@ func hasLocalCommits(branch string, configuration Configuration) bool {
 
 func hasUncommittedChanges() bool {
 	return !isNothingToCommit()
+}
+
+func hasUnpushedCommits(branch string, configuration Configuration) bool {
+	countOutput := silentgit(
+		"rev-list", "--count", "refs/heads/"+branch+"..."+"refs/remotes/"+configuration.RemoteName+"/"+branch,
+	)
+	unpushedCount, err := strconv.Atoi(strings.TrimSpace(countOutput))
+	if err != nil {
+		panic(err)
+	}
+	return unpushedCount != 0
 }
 
 func isMobProgramming(configuration Configuration) bool {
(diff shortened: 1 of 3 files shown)


TrentonAdams:
@simonharrer Oh, i see what you mean.  I think this would work, yes?

```bash
echo hello > blah; git add blah; git commit -m 'blah'
git rev-list --count HEAD...refs/remotes/origin/main
1
```

Then if non zero, there's a local commit that wasn't pushed.

simonharrer:
Yeah, let's try that. Can you introduce a new method that basically only checks that?

TrentonAdams:
I sure can.  I'll see if I can get to that today.

TrentonAdams:
I'm not sure what you wanted to see for line wrapping if any.  Let me know if you want anything changed.

mob.go
```diff
@@ -751,6 +751,17 @@ func hasUncommittedChanges() bool {
 	return !isNothingToCommit()
 }
 
+func hasUnpushedCommits(branch string, configuration Configuration) bool {
+	countOutput := silentgit(
+		"rev-list", "--count", "refs/heads/"+branch+"..."+"refs/remotes/"+configuration.RemoteName+"/"+branch,
+	)
+	unpushedCount, err := strconv.Atoi(strings.TrimSpace(countOutput))
+	if err != nil {
+		panic(err)
```
simonharrer, line 760: an info log about the number of unclpushed commits found would be nice

also in the lastest commit in main the silentgit method now always trims.

simonharrer:
Can you add a changelog entry, please?

simonharrer:
Otherwise looks really great. I already want to say thanks for contributing and helping out making the tool better! :-)

Looking forward to see this enhancement in the next version.

mob.go
TrentonAdams: k, I'll rebase against upstream in a moment.

TrentonAdams:
k, done.  I can rebase to a squashed commit now if you like.  Or if you do that as part of your normal merge that's fine too.

TrentonAdams:
Also, haven't tried from the actual cli yet, just the tests, but I'm going to go figure out how to install locally.  Been a couple months since I learned go stuff. lol

TrentonAdams:
Confirmed...
```
$ ~/go/bin/mob start
   git fetch origin --prune
   git merge FETCH_HEAD --ff-only
 > starting new mob session from origin/main
   git checkout -B mob/main origin/main
   git push --no-verify --set-upstream origin mob/main
 > there are 1 unpushed commits
 ERROR cannot start; unpushed changes on current branch must be pushed upstream
 > you are mob programming
 > on wip branch mob/main (base branch main)
```

TrentonAdams:
Oh, how come it says "you are mob programming on wip branch...".  That seems like an unrelated bug.  Cause we haven't started mob programming if there's an error.

simonharrer:
The other output is something different. It’s something I would like to have a got as well. So if start returns early due to an error it should not call mob status afterwards. Probably needs a separate issue/ticket.

PR #169, opened 2021-05-04
Say "git commit" only on squash

Dosenpfand:
The help text "To finish, use git commit" should only be shown when the commits are squashed. When merging it is not necessary to commit.


simonharrer:
Good catch!

PR #170, opened 2021-05-05
snap

nikolashermann:


README.md
```diff
@@ -69,6 +69,14 @@ On Arch Linux via yay:
 yay -S mobsh-bin
 ```
 
+On Ubuntu via [snap](https://snapcraft.io/mob-sh):
+
+```bash
+sudo snap install mob-sh
+sudo snap alias mob-sh mob
```
simonharrer, line 76: why not name this tool 'mob' from the start? Then, no alias would be necessary.

nikolashermann: Canonical denied my request to register mob, as somehow they don't know abaout the importance of this tool, yet

_Three-letter and shorter snap names are too generic and nondescript and don't provide a good idea of what the snap actually is or does. For that reason, we prefer and suggest you pick a more descriptive name for your snap (mob-handover for example). Keep in mind that even if the snap has one of those names, a command called "mob" can still be provided by the snap - this needs creating an "alias" but it's entirely feasible to do so.

Let me know if some of the suggested names are OK with you or if you want to pick a new one; in any case, you can simply register the new name and it should be automatically approved - only 3-letter or shorter names (well and a few other exceptions) require manual name review/approval."_

simonharrer: :-(

nikolashermann: I requested automatic connection and alias creation to ease to installation process.  
https://forum.snapcraft.io/t/auto-connect-alias-request-for-mob-sh/24303

PR #177, opened 2021-06-09
fix #176 : espeak on Windows WSL2

### LINKED ISSUE #176: eSpeak setup doesn't work with WSL2
it seems that there is no easy access for eSpeak to the sound card in WSL2 on windows.

Workaround is to install eSpeak on windows und use espeak.exe from WSL2

rdmueller:


rdmueller:
I wanted to keep it simple. The windows installer puts espeak in the given location, but it depends on the WSL OS if the path gets mounted to `/mnt/c` or `/c` or even somewhere else. So I thought this owuld be the minimal information someone needs to set it up: a hint not to use the linux espeak but the windows version :-)

Thanx for accepting the PR!

PR #180, opened 2021-06-17
Multiple wip branches

gregorriegler:
This is going in the suggested direction.
What do you think?

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -554,9 +554,9 @@ func start(configuration Configuration) error {
 		return errors.New("cannot start; unpushed changes on base branch must be pushed upstream")
 	}
 
-	wipBranchesWithQualifier := getQualifiedBranches(currentBaseBranch, configuration)
+	wipBranchesWithQualifier := getQualifiedWipBranches(currentBaseBranch, configuration)
 	if !isMobProgramming(configuration) && !hasRemoteBranch(currentWipBranch, configuration) && len(wipBranchesWithQualifier) > 0 && !configuration.WipBranchQualifierSet {
-		sayWithPrefix("other qualified mob branches detected:", " ⚠ ")
+		sayWarning("Creating a new mob branch even though preexisting mob branches have been detected.")
 		for _, wipBranch := range wipBranchesWithQualifier {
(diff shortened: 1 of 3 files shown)


simonharrer:
Goes in the right direction. The warning should only be displayed if it would *create* the new wip branch. Joining an existing wip branch should always work without a warning. The code should reflect that.

The warning itself assumes that the user might perhaps wanted to use a branch modifier. The warning message should reflect that, I think, so the user can fix it if necessary.

There might be another case as well, namely, the inverse: there exists a qualified wip branch (mob/main-green) and one issues `mob start` . There, a message asking the user that perhaps they meant `mob start --branch green` was the one the user wanted.

gregorriegler:
I changed it so that it only shows the warning if the WIP branch does not exist yet.

Would prefer to have the inverse case handled in another PR, what do you think?

mob.go
```diff
@@ -603,11 +602,19 @@ func sayUnstagedChangesInfo() {
 	}
 }
 
-func hasQualifiedBranches(currentBaseBranch string, remoteBranches []string, configuration Configuration) bool {
```
simonharrer, line 606: getQualifiedWipBranches

mob.go
```diff
@@ -563,6 +554,14 @@ func start(configuration Configuration) error {
 		return errors.New("cannot start; unpushed changes on base branch must be pushed upstream")
 	}
 
+	wipBranchesWithQualifier := getQualifiedBranches(currentBaseBranch, configuration)
+	if !isMobProgramming(configuration) && !hasRemoteBranch(currentWipBranch, configuration) && len(wipBranchesWithQualifier) > 0 && !configuration.WipBranchQualifierSet {
+		sayWithPrefix("other qualified mob branches detected:", " ⚠ ")
```
simonharrer, line 559: Explain to user why this warning is shown.

simonharrer:
Please a separate PR for the other change.

Can you add a changelog entry, please

gregorriegler:
improved warnings, added changelog, and extracted sayWarning.

@simonharrer Feel free to improve on the wordings.

PR #186, opened 2021-07-20
Branch value

gregorriegler:
I introduced a Value Object `Branch`.
See where this is going?
Some of the behavior of `Configuration` moved to `Branch` where it is more appropriate.
`Branch` could even accumulate more behavior such as `checkout` and many more.
While it's tempting to move some git wrapper functions into `Branch`, I think this should only be a temporary solution.
At one point we probably want a git interface that abstracts away all git functionality.
Once we have that, git should be passed as an argument to branch.
E.g.
```
git = newGitCliWrapper()
newBranch("mob-session").checkout(git)
```

I see many advantages in this design.

What do you think?

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -625,7 +625,7 @@ func start(configuration Configuration) error {
 	if currentWipBranch.hasRemoteBranch(configuration) {
 		startJoinMobSession(configuration)
 	} else {
-		warnForActiveMobSessions(configuration, currentBaseBranch.String())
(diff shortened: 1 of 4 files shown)


gregorriegler:
**knock knock**

simonharrer:
In general, I like it. I think using silent it okay in those cases. But not sure about the direction regarding the design in combination with the git client. We should discuss this separately.

PR #187, opened 2021-07-24
Improve next mobber suggestion

SimplyDanny:
The existing heuristic which suggests the next mobber is relatively simple. The PR tries to add some improvements, namely:

* Do not suggest the current Git user if there are other mobbers in the list.
* Do not list the name of a previous committer multiple times in a row.
* Apply a simple lookahead in order to also suggest persons which might have been absent in the previous mob round.

Hopefully the tests make my intensions clear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.9.0
 - `mob next` does not show the same committer multiple times in the list of previous committers.
 - `mob next` does not suggest the current Git user to be the next typist as long as there were other persons involved in the mob.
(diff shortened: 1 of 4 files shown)


simonharrer:
Looks great. Thanks for making this simple algorithm better.

Can you please use the wording next typist or next driver instead of next mobber? And we need a changelog entry in the CHANGELOG.md.

I think we can ship this with the next minor release.

SimplyDanny:
Happy to hear that!

Good point, "typist" or "driver" sound better and are even more correct. The changelog has been updated. 👍

PR #206, opened 2021-11-07
style: Remove unnecessary fmt.Sprintf

seanpoulter:
Hi Simon,

This PR resolves two issues that `staticcheck` found and reported in Visual Studio Code. I figured I'd raise this separately so it doesn't get mixed in with any upcoming changes.

Thanks and have a great day!


simonharrer:
Thanks for the contribution! :-)

PR #212, opened 2021-11-11
Aur publish

gregorriegler:
Hi Simon,

This is Sean and me, we think the release.yml is ready for testing.
We deactivated the Homebrew deploy (check line 48).
And we currently added code to output the PKGBUILD file (line 101 to 104).
What you could do is, maybe just create some dummy release to actually test the action.
Maybe we can arrange a meeting, where Sean, you and me can test it together, and adapt accordingly?

Greetings


seanpoulter:
We're ready for your review @draptik and @Morl99. 😁

seanpoulter:
I've tested this on my fork by creating a release without setting up credentials to publish. Here's [the action](https://github.com/seanpoulter/mob/runs/4185763459#step:4:2) in action:

image

<br>

Here is the diff between what we output from the step above and @draptik's version from https://github.com/draptik/arch-package-builds/blob/master/mobsh-bin/PKGBUILD:

image

.github/workflows/release.yaml
```diff
@@ -54,3 +56,61 @@ jobs:
...
+
+          package() {
+              install -D -m644 "LICENSE" "\$pkgdir/usr/share/licenses/\$pkgname/LICENSE"
+              install -D -m755 "mob" "\$pkgdir/usr/bin/mob"
+          }
+
+          EOF
+
+      # this action is just meant for feedback (how does the file look like?)
+      - name: Publish PKGBUILD for ${{ env.PACKAGE_NAME }}
```
Morl99, line 103: You could also store this as an artifact: https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts

.github/workflows/release.yaml
```diff
@@ -54,3 +56,61 @@ jobs:
...
+
+      - name: Publish AUR package
+        uses: KSXGitHub/github-actions-deploy-aur@v2.2.5
+        with:
+          pkgname: ${{ env.PACKAGE_NAME }}
+          pkgbuild: ./aur/${{ env.PACKAGE_NAME }}/PKGBUILD
+          commit_username: ${{ secrets.AUR_USERNAME }}
+          commit_email: ${{ secrets.AUR_EMAIL }}
+          ssh_private_key: ${{ secrets.AUR_SSH_PRIVATE_KEY }}
+          commit_message: Update AUR package
```
Morl99, line 115: Maybe it would be nice to include the version in the message? But I don't knowcif it has any relevance.

.github/workflows/release.yaml
```diff
@@ -43,7 +45,7 @@ jobs:
         run: |
           echo ::set-output name=sha::"$(shasum -a 256 *.tar.gz | awk '{printf $1}')"
       - name: Bump homebrew formula
-        if: matrix.os == 'darwin'
+        if: matrix.os == 'darwin' && 1 == 2 # temporarily disabled
```
Morl99, line 48: This still needs to be removed.

.github/actions/publish-mobsh-bin/action.yaml
```diff
@@ -0,0 +1,51 @@
+name: 'Publish Arch Linux Package - mobsh-bin'
```
Morl99, line 1: This whole file is a duplicate, right?

Morl99 (commented):
This looks very good, I have added a few comments.

simonharrer:
Hm, testing it with a dummy release is a little bit problematic, as anyone installing it during this time via curl will not be able to install it. Can we set up a manual trigger for this action just for testing the latest release?

See https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_dispatch

.github/actions/publish-mobsh-bin/action.yaml
seanpoulter: Oh no! You are correct. Sorry @Morl99.

.github/workflows/release.yaml
seanpoulter: Thanks! That's a great suggestion.

.github/workflows/release.yaml
seanpoulter: Another good suggestion. 👍

seanpoulter:
> Hm, testing it with a dummy release is a little bit problematic, as anyone installing it during this time via curl will not be able to install it. Can we set up a manual trigger for this action just for testing the latest release?
> 
> See https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_dispatch

Have you got manual triggered Workflows to work @simonharrer? I've had troubles with them last week and again this week. What I experience is documented in this thread https://github.community/t/workflow-dispatch-workflow-not-showing-in-actions-tab/130088/15. In order for a new Workflow to appear in the UI like this I have to create a placeholder on `main`:

image

That means to test it like this we will:

- Create `.github/workflows/test-arch-linux-release.yml` on main
- Create `.github/workflows/test-arch-linux-release.yml` on this branch
- Copy and modify the job to use different variable values unless we extract it as an action
- Manually trigger the release using UI

Sound good?

seanpoulter:
@simonharrer , I've set up the Workflow Dispatch Action on [my fork](https://github.com/seanpoulter/mob/actions/workflows/test-arch-linux-release.yml) if you want to see what it looks like. It needed a commit to `main` before I could see or run the workflow. I'd be happy to push the additional commit to the `aur-publish` branch so we get the buggy GitHub UI here too.

Anyways, here's a low quality recording of it in action. From my repo it fails because I don't have your secrets set up. You can also see the `PKGBUILD` file as an artifact that @Morl99 suggested.
https://user-images.githubusercontent.com/2585460/141602814-b017a7cc-f509-4104-9ac2-0d6c5bca32b5.mp4

simonharrer:
Awesome! Thanks for your hard work, already!

We can try out anything in main, that's not an issue for me. So prepare this PR for merging, please, so we can test it on the main branch for the real repository.

.github/workflows/release.yaml
seanpoulter: Done

seanpoulter:
Sounds like a plan Simon. It should be good to go.

Issue #219, opened 2021-12-14
mob done commit behaviour inconsistent, dependent on --no-squash


albran:
**mob version 2.1.0**

the commit behaviour of `mob done` appears to be inconsistent depending if the `--no-squash` flag is included.

without `--no-squash`, `mob done` results in uncommitted, staged changes from the wip branch, requiring a `git commit`
with `--no-squash`, changes on the wip branch are committed to the proper branch, using the default wip commit message

this behaviour does not seem to be specified in the docs, is it intentional?
would be nice to configure `mob done` commit behaviour seperately

gregorriegler:
> The docs specify that --no-squash simply prevents squashing of wip commits assumedly created by mob next commands. 

Ah, I see.
So you probably only want to use `mob done --no-squash` in case you have the commits lined up with correct messages in your wip branch. And you do this by using the `mob squash-wip` command.
So lets say the wip branch has the following commits:

- wip commit
- wip commit
- Refactor so and so
- wip commit
- Feature X done

You may now run `mob squash-wip` and it will get rid of all the wip commits. So you will end up with this on your wip branch:

- Refactor so and so
- Feature X done

After this, when you did `mob done` it would squash everything and remove those commits.
So in stead you do `mob done --no-squash` which will allow you to keep those commits.

I hope this clears up how `mob done --no-squash` came to be.

Now, I'm not sure about `--no-commit` and `--ff` thing. Do you have a better Idea to achieve the thing?

albran:
in the case of our specific workflow, we simply want to keep every single commit from the mob branch when `mob done` does its work. `--no-squash` works fine for this need.

the issue i am trying to raise is the unexpected "side effect" of running `mob done --no-squash` with changes since the previous commit. in this case (as i understand it from examining the console):

1. changes are committed to the mob branch as a WIP commit, using the default WIP commit message
2. in a series of steps, the mob branch is merged into the proper branch using fast-forward (the final merge is executed by `git merge --no-commit --ff mob/my-branch` with the proper branch checked out). in doing so, the proper branch receives the WIP commit message from step 1. as its latest commit. 
3. the mob branch is deleted

from reading the docs, when i execute `mob done` with new changes since the last commit, regardless of `--no-squash`, i expect the opportunity to specify a coherent commit message. in reality, `--no-squash` deprives me of this opportunity, because of what happens in step 2.

gregorriegler:
[...]

PR #221, opened 2021-12-16
test for uncommited files

gregorriegler:


gregorriegler:
This should help us to make our tests more solid by checking the resulting git status.
It would also help us to Test Drive uncommited changes as requested in https://github.com/remotemobprogramming/mob/issues/219

PR #222, opened 2021-12-19
No squash

gregorriegler:
When the user has uncommitted changes and runs `mob done --no-squash`, the changes get commited in a wip commit, but this wip commit gets reset softly after the merge, so the user gets the chance to commit the changes manually with a proper message. 

Tackles this Issue: https://github.com/remotemobprogramming/mob/issues/219

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -1016,10 +1016,10 @@ func done(configuration Configuration) {
 			sayError(err.Error())
 		}
 
-		if isNothingToCommit() {
-			sayInfo("nothing was done, so nothing to commit")
-		} else {
+		if hasUncommittedChanges() {
 			sayTodo("To finish, use", "git commit")
+		} else if configuration.MobDoneSquash {
+			sayInfo("nothing was done, so nothing to commit")
 		}
 
 	} else {
--- a/mob.go
+++ b/mob.go
@@ -985,7 +985,8 @@ func done(configuration Configuration) {
 	baseBranch, wipBranch := determineBranches(gitCurrentBranch(), gitBranches(), configuration)
 
 	if wipBranch.hasRemoteBranch(configuration) {
-		if hasUncommittedChanges() {
+		uncommittedChanges := hasUncommittedChanges()
+		if uncommittedChanges {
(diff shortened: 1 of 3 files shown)


simonharrer:
Can you add a CHANGELOG entry? Otherwise, I think this is reasonable.

mob.go
```diff
@@ -998,6 +999,11 @@ func done(configuration Configuration) {
 		}
 
 		git("branch", "-D", wipBranch.Name)
+
+		if uncommittedChanges && !configuration.MobDoneSquash { // give the user the chance to name their final commit
+			silentgit("reset", "--soft", "HEAD^")
```
simonharrer, line 1004: I think this should not be a silentgit but a git as this is helpful for the user to see.

albran (changes requested):
at `1018` in `mob.go`
```
if configuration.MobDoneSquash {
	if isNothingToCommit() {
		sayInfo("nothing was done, so nothing to commit")
	} else {
		sayTodo("To finish, use", "git commit")
	}
}
```

`if configuration.MobDoneSquash` could be removed to receive the same todo prompt when specifying ```--no-squash```

gregorriegler:
@simonharrer @albran 
Thank you for your inputs. Made the changes.

gregorriegler:
@simonharrer should I merge it?

mob.go
```diff
@@ -1009,12 +1015,11 @@ func done(configuration Configuration) {
...
 		}
-		if configuration.MobDoneSquash {
-			if isNothingToCommit() {
-				sayInfo("nothing was done, so nothing to commit")
-			} else {
-				sayTodo("To finish, use", "git commit")
-			}
+
+		if isNothingToCommit() {
+			sayInfo("nothing was done, so nothing to commit")
```
simonharrer, line 1020: When doing a merge, this message is wrong. We need two different messages, one for each option.

simonharrer: In case of `--no-squash` we could basically omit it. That would be fine as well.

gregorriegler:
@simonharrer you mean like this?

PR #235, opened 2022-02-13
Reliable audio with embedded mp3 files

simonharrer:
Explorative try to more reliable audio by:
- using beep library
- embedding mp3 files (downside: no longer real text to speech, requires go 1.16)
- forked sleep timer command calls `mob play mobnext` to play the audio 

TODOs:
- get the tests working again
- get the pipeline running for every platform so that audio works
- double check error handling of sound playing library

Ideas:
- use "next" instead of "mob next" for more neutral audio output
- use "start" instead of "mob start" for more neutral audio output

PR #236, opened 2022-02-18
feat: Add custom next sound possibility

lizell:
Adds the possibility to install a custom next sound to be played instead of the voice command "mob next".

This is an alternative to #235. Feel free to close this, if you want to continue the 235 way.

TODO?:
* Add support for linux/windows (only mac support currently)


gregorriegler:
Thank you lizell.
I have a few of concerns with this solution.
A custom sound is a thing you would want to configure in your configuration file.
And it should be well encapsulated and work on all OS.
I will close this for now. Feel free to reach out if you're still interested in thinking this through.

PR #237, opened 2022-02-19
Squash wip improvement

gregorriegler:
Adds new `mob done --squash-wip` option, which does a `mob done` but also runs `squashWip` beforehand.
As a result, when you use `mob done --squash-wip`, you can keep the manual commits you made on the wip branch while getting rid of all the wip commits.

`squashWip` had to be tweaked for this to fixup subsequent wip commits that had no following manual commit. And it also undoes a final wip commit, so that the user is forced to commit those changes manually.

The environment var MOB_DONE_SQUASH is made backward compatible, while the `.mob` config file does not support activating --squash-wip yet.

Questions:

- Should we now already discard the `mob squash-wip` command after all?
- Can you find any corner cases I missed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ Basic Commands(Options):
   done
     [--no-squash]                        Do not squash commits from wip branch
     [--squash]                           Squash commits from wip branch
+    [--squash-wip]                       Squash wip commits from wip branch, maintaining manual commits
   reset
     [--branch|-b <branch-postfix>]       Set wip branch to 'mob/<base-branch>-<branch-postfix>'
 
--- a/README.md
+++ b/README.md
@@ -174,7 +174,8 @@ Basic Commands(Options):
 
 Experimental Commands:
   squash-wip                             Combines wip commits in wip branch with subsequent manual commits to leave only manual commits.
-                                         ! Works only if all wip commits have the same wip commit message !
+                                         !Deprecated! As of version 2.6.0 this command is obsolete as the behaviour is added to 'done --squash-wip' which is more convenient.
+                                         
     [--git-editor]                       Not intended for manual use. Used as a non-interactive editor (GIT_EDITOR) for git.
     [--git-sequence-editor]              Not intended for manual use. Used as a non-interactive sequence editor (GIT_SEQUENCE_EDITOR) for git.
 
--- a/README.md
+++ b/README.md
@@ -172,13 +172,6 @@ Basic Commands(Options):
   reset
     [--branch|-b <branch-postfix>]       Set wip branch to 'mob/<base-branch>-<branch-postfix>'
 
-Experimental Commands:
-  squash-wip                             Combines wip commits in wip branch with subsequent manual commits to leave only manual commits.
-                                         !Deprecated! As of version 2.6.0 this command is obsolete as the behaviour is added to 'done --squash-wip' which is more convenient.
-                                         
-    [--git-editor]                       Not intended for manual use. Used as a non-interactive editor (GIT_EDITOR) for git.
-    [--git-sequence-editor]              Not intended for manual use. Used as a non-interactive sequence editor (GIT_SEQUENCE_EDITOR) for git.
-
 Timer Commands:
   timer <minutes>    start a <minutes> timer
   start <minutes>    start mob session in wip branch and a <minutes> timer
--- a/mob.go
+++ b/mob.go
@@ -667,6 +667,8 @@ func parseArgs(args []string, configuration Configuration) (command string, para
 			newConfiguration.DoneSquash = Squash
 		case "--no-squash":
 			newConfiguration.DoneSquash = NoSquash
+		case "--squash-wip":
+			newConfiguration.DoneSquash = SquashWip
 		default:
 			if i == 1 {
 				command = arg
@@ -1424,6 +1426,7 @@ Basic Commands(Options):
   done
     [--no-squash]                        Do not squash commits from wip branch
     [--squash]                           Squash commits from wip branch
+    [--squash-wip]                       Squash wip commits from wip branch, maintaining manual commits
   reset
     [--branch|-b <branch-postfix>]       Set wip branch to 'mob/<base-branch>/<branch-postfix>'
 
--- a/mob.go
+++ b/mob.go
@@ -31,9 +31,21 @@ type DoneSquash string
 
 const (
 	Squash    DoneSquash = "squash"
-	NoSquash  DoneSquash = "no-squash"
-	SquashWip DoneSquash = "squash-wip"
+	NoSquash             = "no-squash"
+	SquashWip            = "squash-wip"
 )
+
+func doneSquash(value string) DoneSquash {
+	switch value {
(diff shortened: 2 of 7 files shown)


hollesse:
It seems to me that the adjusted configuration option squash done can not be set from the configuration files. I just saw the parsing of a Boolean and not a string, like you did it for the parsing of the environment variables.

gregorriegler:
> It seems to me that the adjusted configuration option squash done can not be set from the configuration files. I just saw the parsing of a Boolean and not a string, like you did it for the parsing of the environment variables.

Correct

mob.go
```diff
@@ -467,6 +475,20 @@ func setBoolean(s *bool, key string, value string) {
 	debugInfo("Overwriting " + key + " =" + strconv.FormatBool(boolValue))
 }
 
+func setMobDoneSquash(configuration *Configuration, key string, value string) {
```
simonharrer, line 478: Hm, not sure whether we really benefit from the enum here. Just using a string would probably be simpler.

squash_wip_test.go
```diff
@@ -297,3 +346,31 @@ squash c51a56d manual commit
 	result, _ := ioutil.ReadFile(input)
 	equals(t, expected, string(result))
 }
+
+func wipCommit(t *testing.T, configuration Configuration, filename string) {
+	start(configuration)
+	createFile(t, filename, "irrelevant")
```
simonharrer, line 352: could you use `"contentIrrelevant"` in all cases? Little bit easier to read in github PRs.

squash_wip.go
```diff
@@ -32,13 +24,24 @@ last commit must be a manual commit`)
...
+	sayInfo("the history of your '" + currentWipBranch.String() + "' branch has been rewritten to combine all wip commits, if possible with their following manual commits:")
 	sayEmptyLine()
 	sayLastCommitsWithMessage(currentBaseBranch.String(), currentWipBranch.String())
-	sayEmptyLine()
-	sayTodo("to finally put the changes into the base branch preserving the resulting commits, call:", configuration.mob("done --no-squash"))
+	if lastCommitIsWipCommit(configuration) { // last commit is wip commit
+		sayInfo("undoing the final wip commit and staging its changes:")
+		git("reset", "--soft", "HEAD^")
+	}
+	git("push", "--force", "--no-verify")
```
simonharrer, line 36: You need to get the info from the configuration, see `configuration.gitHooksOption()`

squash_wip.go
```diff
@@ -32,13 +24,24 @@ last commit must be a manual commit`)
 		mobExecutable()+" squash-wip --git-editor",
 		mobExecutable()+" squash-wip --git-sequence-editor",
 	)
-	silentgit("rebase", "-i", "--keep-empty", mergeBase)
+	git("rebase", "-i", "--keep-empty", mergeBase)
```
simonharrer, line 27: having a sayInfo describing what is about to happen before the rebase might be helpful

squash_wip.go
```diff
@@ -32,13 +24,24 @@ last commit must be a manual commit`)
...
 	setEnvGitEditor(originalGitEditor, originalGitSequenceEditor)
-	sayInfo("the history of your '" + currentWipBranch.String() + "' branch has been rewritten to combine all wip commits with their following manual commits:")
+	sayInfo("the history of your '" + currentWipBranch.String() + "' branch has been rewritten to combine all wip commits, if possible with their following manual commits:")
```
simonharrer, line 29: I don't like this "if possible", but I don't have a great idea at hand...

squash_wip.go
simonharrer, line 27: Instead of "-i" we should write "--interactive" (always use the more readable parameter)

mob.go
gregorriegler: It is a string. It's just masked with its own type. This helps with refactoring. Also, it makes sure that we don't accidentally use invalid values, so it isolates this invariant. And on top of that, it helps me with coding, as I don't have to remember what exactly the string was, I just use auto-completion. I understand, that the translation code from env variable to this new Type is a little complicated. But it would be as complicated if it was a string. So if you asked me, I'm all for the enumish Type wrapping.

squash_wip_test.go
gregorriegler: I agree

mob.go
gregorriegler: I just improved on the code that parses the DoneSquash configuration. Its odd. Even though they are just strings, I cannot compare them to strings.

mob.go
```diff
@@ -27,29 +27,49 @@ var (
...
+const (
+	Squash    DoneSquash = "squash"
+	NoSquash             = "no-squash"
+	SquashWip            = "squash-wip"
+)
+
+func doneSquash(value string) DoneSquash {
+	switch value {
+	case "false":
+		fallthrough
```
simonharrer, line 41: you can have two cases comma separated...

simonharrer:
Can you add a changelog entry for the next version?

squash_wip.go
```diff
@@ -32,13 +24,25 @@ last commit must be a manual commit`)
 		mobExecutable()+" squash-wip --git-editor",
 		mobExecutable()+" squash-wip --git-sequence-editor",
 	)
-	silentgit("rebase", "-i", "--keep-empty", mergeBase)
+	sayInfo("rewriting the history of the '" + currentWipBranch.String() + "' branch to squash wip commits but keep manual commits.")
```
simonharrer, line 27: shorter wording: `rewriting history of %s: squashing wip commits while keeping manual commits`

simonharrer:
LGTM

gregorriegler:
I did the final two changes from my Phone 😅.
Can't do the changelog entry right now. If you want it now, feel free to add it.

gregorriegler:
All set

gregorriegler:
Ok, I just added a note that the old squash-wip command is deprecated.

README.md
```diff
@@ -174,7 +174,8 @@ Basic Commands(Options):
 
 Experimental Commands:
   squash-wip                             Combines wip commits in wip branch with subsequent manual commits to leave only manual commits.
-                                         ! Works only if all wip commits have the same wip commit message !
+                                         !Deprecated! As of version 2.6.0 this command is obsolete as the behaviour is added to 'done --squash-wip' which is more convenient.
+                                         
     [--git-editor]                       Not intended for manual use. Used as a non-interactive editor (GIT_EDITOR) for git.
     [--git-sequence-editor]              Not intended for manual use. Used as a non-interactive sequence editor (GIT_SEQUENCE_EDITOR) for git.
```
simonharrer, line 180: Would remove this alltogether. We don't have to explain when removing experimental commands.

simonharrer:
We should remove the squash-wip command, or add a deprecation warning when this command is called.

gregorriegler:
I was not able to remove the command entirely, as we use it as an editor in the squash-wip feature. But I left only the editor part and removed it from the docs.

simonharrer:
Awesome. Thanks for this great PR! :-)

PR #241, opened 2022-03-08
Add smiley face to Happy Collaborating message

aryanfaghihi:
How can one be happy collaborating without seeing a smiley face? 

This PR simply adds that smiley face to make your day :)

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -702,7 +702,7 @@ func execute(command string, parameter []string, configuration Configuration) {
 		} else if configuration.Timer != "" {
 			startTimer(configuration.Timer, configuration)
 		} else {
-			sayInfo("It's now " + currentTime() + ". Happy collaborating!")
+			sayInfo("It's now " + currentTime() + ". Happy collaborating! :)")
 		}
 	case "b", "branch":
 		branch(configuration)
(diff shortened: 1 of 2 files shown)


mob.go
```diff
@@ -702,7 +702,7 @@ func execute(command string, parameter []string, configuration Configuration) {
 		} else if configuration.Timer != "" {
 			startTimer(configuration.Timer, configuration)
 		} else {
-			sayInfo("It's now " + currentTime() + ". Happy collaborating!")
+			sayInfo("It's now " + currentTime() + ". Happy collaborating! :)")
```
ngocdaothanh, line 705: Should we use the Unicode one?
https://emojipedia.org/smiling-face-with-smiling-eyes/

aryanfaghihi: I kind of like the subtle nature of ":)" but open to using a Unicode one too :)

simonharrer:
Great PR :)

Because it is so great, we need a changelog entry. Can you add that, please?

mob.go
simonharrer: I would go with the subtle nature for now.

aryanfaghihi:
@simonharrer No worries, done!

PR #245, opened 2022-03-17
Open last modified file of previous typist

hollesse:
One thing I was thinking about is, does it make sense to set defaults here or should I just print a warning and do nothing if the open command is not set.


squash_wip.go
```diff
@@ -186,7 +186,7 @@ func isManualCommit(line string, configuration Configuration) bool {
 }
 
 func isWipCommit(line string, configuration Configuration) bool {
-	return strings.HasSuffix(line, configuration.WipCommitMessage)
+	return strings.Contains(line, configuration.WipCommitMessage)
```
simonharrer, line 189: why contains and not hasPrefix similarly to the change above?

hollesse: As you can see, the function used has suffix before because there is a abbreviated commit hash in Front on the commit message. I am not sure if this hash is needed of if we could change the code that we could just get the message here.

mob.go
```diff
@@ -45,28 +45,31 @@ func doneSquash(value string) string {
...
+	NotifyCommand                        string // override with MOB_NOTIFY_COMMAND
+	NotifyMessage                        string // override with MOB_NOTIFY_MESSAGE
+	NextStay                             bool   // override with MOB_NEXT_STAY
+	StartIncludeUncommittedChanges       bool   // override with MOB_START_INCLUDE_UNCOMMITTED_CHANGES variable
+	StashName                            string // override with MOB_STASH_NAME
+	WipBranchQualifier                   string // override with MOB_WIP_BRANCH_QUALIFIER
+	WipBranchQualifierSeparator          string // override with MOB_WIP_BRANCH_QUALIFIER_SEPARATOR
+	WipBranchPrefix                      string // override with MOB_WIP_BRANCH_PREFIX
+	DoneSquash                           string // override with MOB_DONE_SQUASH
+	WriteLastModifiedFileInCommitMessage bool   // override with MOB_WRITE_LAST_MODIFIED_FILE_IN_COMMIT_MESSAGE
```
simonharrer, line 64: Why do we need a flag for this? Why not always do this, if possible?

mob.go
```diff
@@ -264,38 +267,46 @@ func currentCliName(argZero string) string {
...
 		voiceCommand = "say \"%s\""
 		notifyCommand = "/usr/bin/osascript -e 'display notification \"%s\"'"
+		openCommand = "open -e %s"
 	case "linux":
 		voiceCommand = "say \"%s\""
 		notifyCommand = "notify-send \"%s\""
+		openCommand = "editor %s"
 	case "windows":
 		voiceCommand = "(New-Object -ComObject SAPI.SPVoice).Speak(\\\"%s\\\")"
+		openCommand = "Notepad \"%s\""
```
simonharrer, line 282: I don't think default values for this open command make sense. This is something everybody has to set explicitly themselves.

mob.go
hollesse: I'm not sure if we would get any problems, if one has the new version of mob with this feature and one has an old version. If the one with the old version does a mod done, it maybe could produce problems while squashing, because his mob version would not recognize the commit as a WIP commit.

mob.go
```diff
@@ -45,28 +45,31 @@ func doneSquash(value string) string {
...
+	NextStay                             bool   // override with MOB_NEXT_STAY
+	StartIncludeUncommittedChanges       bool   // override with MOB_START_INCLUDE_UNCOMMITTED_CHANGES variable
+	StashName                            string // override with MOB_STASH_NAME
+	WipBranchQualifier                   string // override with MOB_WIP_BRANCH_QUALIFIER
+	WipBranchQualifierSeparator          string // override with MOB_WIP_BRANCH_QUALIFIER_SEPARATOR
+	WipBranchPrefix                      string // override with MOB_WIP_BRANCH_PREFIX
+	DoneSquash                           string // override with MOB_DONE_SQUASH
+	WriteLastModifiedFileInCommitMessage bool   // override with MOB_WRITE_LAST_MODIFIED_FILE_IN_COMMIT_MESSAGE
+	OpenLastModifiedFile                 bool   // override with MOB_OPEN_LAST_MODIFIED_FILE
+	OpenCommand                          string // override with MOB_OPEN_COMMAND
```
simonharrer, line 66: I think setting the MOB_OPEN_COMMAND to a nonempty string should enable the open feature.

mob.go
hollesse: I also thought about this. I agree we should check if this command is set and if not we should show a warning.

mob.go
hollesse: Sounds like a good idea :)

mob.go
```diff
@@ -1063,9 +1098,49 @@ func start(configuration Configuration) error {
...
 
+	if configuration.OpenLastModifiedFile{
+		openLastModifiedFileIfPresent(configuration)
+	}
+
 	return nil // no error
 }
 
+func openLastModifiedFileIfPresent(configuration Configuration) {
+	debugInfo("OpenLastModifiedFile1")
```
simonharrer, line 1109: These debugInfos I would remove before release.

mob.go
```diff
@@ -1063,9 +1098,49 @@ func start(configuration Configuration) error {
...
+	if !lastCommitIsWipCommit(configuration){
+		return;
```
simonharrer, line 1111: I would add a debugInfo here

mob.go
```diff
@@ -1063,9 +1098,49 @@ func start(configuration Configuration) error {
...
+	}
+	if len(split) > 2 {
+		sayWarning("Could not determine last modified file from commit message, separator was used multiple times!")
+		return;
+	}
+	debugInfo("OpenLastModifiedFile3")
+	lastModifiedFile := gitRootDir() + "/" + split[1]
+	debugInfo("Last modified file: "+lastModifiedFile)
+	if lastModifiedFile == ""{
+		return;
```
simonharrer, line 1128: I would add a debugInfo here

mob.go
```diff
@@ -1205,11 +1280,62 @@ func getCachedChanges() string {
 
 func makeWipCommit(configuration Configuration) {
 	git("add", "--all")
-	gitWithoutEmptyStrings("commit", "--message", configuration.WipCommitMessage, configuration.gitHooksOption())
+	commitMessage := configuration.WipCommitMessage
+	if configuration.WriteLastModifiedFileInCommitMessage && areFilesModified() {
+		commitMessage += " lastFile:" + getPathOfLastModifiedFile()
+	}
```
simonharrer, line 1286: extract to createWipCommitMessage

mob.go
```diff
@@ -1205,11 +1280,62 @@ func getCachedChanges() string {
...
+			lastModifiedTime = info.ModTime()
+			lastModifiedFilePath = filepath
+		}
+		debugInfo(info.ModTime().String())
+	}
+	return lastModifiedFilePath
+}
+
+// uses git status --short. To work properly files have to be staged.
+func areFilesModified() bool {
```
simonharrer, line 1327: Why do we need this method? We could just use the previous one, and if nothing is found, we don't add the last modified file.

mob.go
```diff
@@ -1205,11 +1280,62 @@ func getCachedChanges() string {
...
+		if strings.HasPrefix(line, "M")  {
+			filepath = strings.TrimPrefix(line, "M")
+		} else if strings.HasPrefix(line, "A") {
+			filepath = strings.TrimPrefix(line, "A")
+		}
+		filepath = strings.TrimSpace(filepath)
+		debugInfo(filepath)
+		info, err := os.Stat(rootDir + "/" + filepath)
+		if err != nil {
+			panic(err)
```
simonharrer, line 1315: Return an error or empty string, or both.

mob.go
```diff
@@ -1205,11 +1280,62 @@ func getCachedChanges() string {
...
+	for _, line := range lines {
+		if strings.HasPrefix(line, "D") || strings.HasPrefix(line, "R"){
+			continue
+		}
+		filepath := ""
+		if strings.HasPrefix(line, "M")  {
+			filepath = strings.TrimPrefix(line, "M")
+		} else if strings.HasPrefix(line, "A") {
+			filepath = strings.TrimPrefix(line, "A")
+		}
```
simonharrer, line 1310: Why not add an else with a continue here? Then you wouldn't need the guard at the beginning?

mob.go
```diff
@@ -1205,11 +1280,62 @@ func getCachedChanges() string {
...
+		filepath = strings.TrimSpace(filepath)
+		debugInfo(filepath)
+		info, err := os.Stat(rootDir + "/" + filepath)
+		if err != nil {
+			panic(err)
+		}
+		if info.ModTime().After(lastModifiedTime){
+			lastModifiedTime = info.ModTime()
```
simonharrer, line 1318: extract `info.ModTime()` to variable `currentLastModifiedTime` so not to call it twice.

mob.go
simonharrer: Hm, I would take that into account.

mob.go
hollesse: done

mob.go
gregorriegler: I find these kind of checks scary. It means that we're handling this on a very low level, but everywhere. The invariant here is that it will always be either an empty string, or a specific string, but never anything in between. I would want to isolate this invariant in its own type. So that we abstract the low level detail away from the rest of the code. But that's just how I see it.

mob.go
```diff
@@ -1063,9 +1086,49 @@ func start(configuration Configuration) error {
 	sayInfo("you are on wip branch '" + currentWipBranch.String() + "' (base branch '" + currentBaseBranch.String() + "')")
 	sayLastCommitsList(currentBaseBranch.String(), currentWipBranch.String())
 
+	if configuration.OpenCommand != "" {
```
gregorriegler, line 1089: I find these kind of checks scary. It means that we're handling this on a very low level, but everywhere. The invariant here is that it will always be either an empty string, or a specific string, but never anything in between. I would want to isolate this invariant in its own type. So that we abstract the low level detail away from the rest of the code. But that's just how I see it.

mob.go
hollesse: done

mob.go
hollesse: done.

mob.go
hollesse: done.

mob.go
hollesse: done

mob.go
hollesse: just removed it.

mob.go
hollesse: added an Error Message here and continued.

mob.go
hollesse: done.

mob.go
hollesse: What do you mean with a specific string here? The string can be any type of command which opens a file. For example you could set it to `idea %s` if you want to use IntelliJ, but it could also be `atom %s`, `code %s` or any other IDE or texteditor you can image. So I don't think we should specify how this open command can be set.

gregorriegler:
The Test TestStartNextStay_OpenLastModifiedFile fails on my system
```
=== RUN   TestStartNextStay_OpenLastModifiedFile
    mob_test.go:1395: Creating testbed in temporary directory /tmp/TestStartNextStay_OpenLastModifiedFile2544377890/001
    mob_test.go:1395: 
        ===== cd /tmp/TestStartNextStay_OpenLastModifiedFile2544377890/001/local
    mob_test.go:1395:   git fetch origin --prune
    mob_test.go:1395:   git merge FETCH_HEAD --ff-only
    mob_test.go:1395: > starting new session from origin/master
    mob_test.go:1395:   git checkout -B mob-session origin/master
    mob_test.go:1395:   git push --no-verify --set-upstream origin mob-session
    mob_test.go:1395: > you are on wip branch 'mob-session' (base branch 'master')
    mob_test.go:1395:   git add --all
    mob_test.go:1395:   git commit --message mob next [ci-skip] [ci skip] [skip ci] lastFile:file.txt --no-verify
    mob_test.go:1395:     file.txt | 1 +
    mob_test.go:1395:     1 file changed, 1 insertion(+)
    mob_test.go:1395:     a9c7e6c4fcaa6f3797efecfff1539857e9a484cd
    mob_test.go:1395:   git push --no-verify origin mob-session
    mob_test.go:1395:   git fetch origin --prune
    mob_test.go:1395: > joining existing session from origin/mob-session
    mob_test.go:1395:   git checkout -B mob-session origin/mob-session
    mob_test.go:1395:   git branch --set-upstream-to=origin/mob-session mob-session
    mob_test.go:1395: > you are on wip branch 'mob-session' (base branch 'master')
    mob_test.go:1395: a9c7e6c 0 seconds ago <local>
    mob_test.go:1537: goroutine 19 [running]:
        runtime/debug.Stack()
        	/usr/lib/go/src/runtime/debug/stack.go:24 +0x65
        github.com/remotemobprogramming/mob/v2.equals(0xc00008ba00, {0x6afa00, 0xc000300990}, {0x6afa00, 0xc00018d110})
        	/home/gregor/IdeaProjects/mob/mob_test.go:1537 +0x4e
        github.com/remotemobprogramming/mob/v2.assertGitStatus(0x6afa00, 0xc000300990)
        	/home/gregor/IdeaProjects/mob/mob_test.go:1523 +0x45
        github.com/remotemobprogramming/mob/v2.TestStartNextStay_OpenLastModifiedFile(0x0)
        	/home/gregor/IdeaProjects/mob/mob_test.go:770 +0x1e9
        testing.tRunner(0xc00008ba00, 0x712830)
        	/usr/lib/go/src/testing/testing.go:1259 +0x102
        created by testing.(*T).Run
        	/usr/lib/go/src/testing/testing.go:1306 +0x35a
        
mob_test.go:1538:

	exp: map[string]string{"file.txt-1":"??"}

	got: map[string]string{}

    testing.go:967: TempDir RemoveAll cleanup: unlinkat /tmp/TestStartNextStay_OpenLastModifiedFile2544377890/001/local: directory not empty
--- FAIL: TestStartNextStay_OpenLastModifiedFile (0.16s)

FAIL

Process finished with the exit code 1


```

mob.go
gregorriegler: I made some commits, tell me what you think. Now I wonder, if we need the `configuration.isOpenCommandGiven` at all, or we can just use `configuration.openCommandFor` which would be even better.

gregorriegler:
I made a commit that fixes the test... But I don't like how it does it. It seems like I have to wait for touch to finish.

mob.go
```diff
@@ -1063,9 +1090,45 @@ func start(configuration Configuration) error {
...
 	sayLastCommitsList(currentBaseBranch.String(), currentWipBranch.String())
 
+	openLastModifiedFileIfPresent(configuration)
+
 	return nil // no error
 }
 
+func openLastModifiedFileIfPresent(configuration Configuration) {
+	if !configuration.isOpenCommandGiven() {
+		return
```
simonharrer, line 1100: would add a debugInfo here as well

simonharrer:
Regarding the wip commit message: 

Currently, it's so:

```
mob next [ci-skip] [ci skip] [skip ci] lastFile: index.html
```

Perhaps the following would be better? In case we want to add more meta data, it's simply one meta data thing per line.

```
mob next [ci-skip] [ci skip] [skip ci]

lastFile: index.html
```

hollesse:
I also had that in mind first, but that does not work with the current implementation of the squashing. Because of this I choose the easier way and just append it without a new line.

mob.go
hollesse: I thought about this the last night. We should be really careful. If we would break the squashing feature of older version. Maybe we should mark this feature as experimental first and let people activate it, but then adjust the squashing. After a few releases we could remove the option to activate it and activate it by default. Then we would not just break the squashing if one of the team uses one version lower then the others. What do you think about that?

simonharrer: I think not a lot of users use the squashing feature. That’s why I don’t think this migration path is necessary. We can increase the major version to signal the major change.

simonharrer:
I would go the extra mile and change the squashing code.

hollesse:
i just added the shortcut if just one file was modified.

mob.go
hollesse: done

hollesse:
@gregorriegler your changes look good to me :). Regarding the tests. I don't have a better idea. So if it works, its okay for me to wait a moment.

mob.go
hollesse: ok

hollesse:
> Regarding the wip commit message:
> 
> Currently, it's so:
> 
> ```
> mob next [ci-skip] [ci skip] [skip ci] lastFile: index.html
> ```
> 
> Perhaps the following would be better? In case we want to add more meta data, it's simply one meta data thing per line.
> 
> ```
> mob next [ci-skip] [ci skip] [skip ci]
> 
> lastFile: index.html
> ```

Done. You can have a look at the changes again @simonharrer @gregorriegler

simonharrer:
Can you update the README.md, please? It no longer matches the actual configuration.

mob.go
```diff
@@ -1205,11 +1261,78 @@ func getCachedChanges() string {
...
 	sayInfoIndented(getChangesOfLastCommit())
 	sayInfoIndented(gitCommitHash())
 }
 
+func createWipCommitMessage(configuration Configuration) string {
+	commitMessage := configuration.WipCommitMessage
+
+	lastModifiedFilePath := getPathOfLastModifiedFile()
+	if lastModifiedFilePath != "" {
+		commitMessage += "\nlastFile:" + lastModifiedFilePath
```
simonharrer, line 1275: Shouldn't it be two new lines? Because in the typical git commit message layout, you have the first line, and then an empty line, and then the other info.

squash_wip_test.go
```diff
@@ -141,7 +141,7 @@ func TestCommitsOnCurrentBranch(t *testing.T) {
 	commits := commitsOnCurrentBranch(configuration)
 
 	equals(t, []string{
-		configuration.WipCommitMessage,
+		configuration.WipCommitMessage + " lastFile:file3.txt",
```
simonharrer, line 144: Hm, I think the test here should fail because of the different commit message. Why does it not fail?

mob.go
hollesse: mh. I can change this, but then I have to change the logic of the squashing again. But I think I should change that anyways.

squash_wip_test.go
hollesse: This returned just the subject. If you have a subject with one linebreak this gets replaced with an whitespace by git. Now the lastFile is not part of the subject anymore, so this can be removed.

mob.go
hollesse: done

hollesse:
> Can you update the README.md, please? It no longer matches the actual configuration.

done.

squash_wip.go
```diff
@@ -114,15 +114,16 @@ func replaceFileContents(fileName string, replacer Replacer) {
...
-	for _, line := range strings.Split(input, "\n") {
+	ignoreBlock := false
+	lines := strings.Split(input, "\n")
+	for idx, line := range lines {
 		if configuration.isWipCommitMessage(line) {
-			isWipCommitBlock = true
-		} else if line == "" {
-			isWipCommitBlock = false
+			ignoreBlock = true
+		} else if line == "" && len(lines) > idx+1 && strings.HasPrefix(lines[idx+1], "#") {
```
gregorriegler, line 122: This condition is hard to make sense of. I think we should move it in it's own function with a descriptive name.

README.md
```diff
@@ -348,6 +355,9 @@ MOB_WIP_BRANCH_QUALIFIER=""
 MOB_WIP_BRANCH_QUALIFIER_SEPARATOR="-"
 MOB_WIP_BRANCH_PREFIX="mob/"
 MOB_DONE_SQUASH=true
+MOB_WRITE_LAST_FILE_IN_COMMIT_MESSAGE=false
+MOB_OPEN_LAST_MODIFIED_FILE=false
```
simonharrer, line 359: those two options need to be removed

squash_wip.go
hollesse: please have a look at the newly extracted function isNextLineComment. Is that fine for you?

README.md
hollesse: done.

PR #249, opened 2022-03-30
first try at the clean command

### LINKED ISSUE #246: What to do when someone else executed mob done?
Given the use case from the README:

```
# Carola
main $ mob start
mob/main $ echo "hello" > work.txt
mob/main $ mob next

# Maria
main $ mob start
mob/main $ cat work.txt # shows "hello"
mob/main $ echo " world" >> work.txt
mob/main $ mob next

# Carola
mob/main $ mob start
mob/main $ cat work.txt # shows "hello world"
mob/main $ echo "!" >> work.txt
mob/main $ mob done
main $ git commit -m "create greeting file"
main $ git push

# Maria
# ???
```

What command does Maria need to execute now? She is still on the mob branch, but wants to go back to the main branch (and probably remove her local mob branch).

* We usually do a `git switch main`. But this does not feel right and leaves a local mob branch laying around
* `mob reset` looks kind of dangerous since it removes the remote branch. So if there were some communication issues, we might remove a branch that someone might still need

Maybe there could be a  paragraph in the README which explains this. Or should there be a new command?

simonharrer:
Fixes #246

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -785,9 +785,9 @@ func clean(configuration Configuration) {
 		if currentBaseBranch.exists(localBranches) {
 			git("checkout", currentBaseBranch.Name)
 		} else if newBranch("main").exists(localBranches) {
-			git("checkout", newBranch("main").Name)
+			git("checkout", "main")
 		} else {
-			git("checkout", newBranch("master").Name)
+			git("checkout", "master")
 		}
 	}
 
--- a/mob.go
+++ b/mob.go
@@ -778,23 +778,30 @@ func clean(configuration Configuration) {
 	currentBranch := gitCurrentBranch()
 	localBranches := gitBranches()
 
-	if currentBranch.isStale(configuration) {
-		currentBaseBranch, currentWipBranch := determineBranches(currentBranch, localBranches, configuration)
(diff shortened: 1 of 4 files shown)


hollesse:
This feature will be most likely used if the team uses feature branches. What if the feature branch is already merged and locally deleted. But the WIP Branch still exists and the user is on the WIP Branch and executes mob clean. The switch to the base branch would fail. Is that correct? In my opinion we should then maybe change to main or master branch.

mob.go
```diff
@@ -770,6 +772,32 @@ func execute(command string, parameter []string, configuration Configuration) {
...
+		b := newBranch(branch)
+		if b.isStale(configuration) {
+			sayInfo("Removing stale wip branch " + b.Name)
+			git("branch", "-d", b.Name)
+		}
+	}
+
+}
+
+func (branch Branch) isStale(configuration Configuration) bool {
```
hollesse, line 797: What do you think about the function name isOrphan? I needed to Google stale. I think orphan could be a better wording because to branch has lost it’s remote origin.

simonharrer: Yes, makes sense.

hollesse:
Looks good to me. Would maybe add two more tests, one with an orphan and an existing remote base branch which is not main or master. And one test with an mob branch which is not an orphan.

simonharrer:
@hollesse done

PR #254, opened 2022-04-11
feature: allow user to retain the WIP branches after `mob done`

hihithisisme:
Hi new user here wanting to suggest a new feature. 
My team has recently adopted the mob tool as part of our process. After drafting a Pull Request for the larger team, we would sometimes have to revisit the WIP branch to continue iteration on the changes. 

Although it's not too much trouble to just branch out again, we thought this would be good to allow the user to explicitly choose if he wants to retain the WIP branch while running `mob done`. 

I went ahead to draft a PR out to show what I mean. Thanks for taking a look!

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ Basic Commands(Options):
     [--no-squash]                        Squash no commits from wip branch, only merge wip branch
     [--squash]                           Squash all commits from wip branch
     [--squash-wip]                       Squash wip commits from wip branch, maintaining manual commits
+    [--retain]							 Prevent the local and remote wip branches from being deleted
   reset
     [--branch|-b <branch-postfix>]       Set wip branch to 'mob/<base-branch>/<branch-postfix>'
   clean                                  Removes all orphan wip branches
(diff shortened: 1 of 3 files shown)


simonharrer:
Thanks for reaching out. I'd like to learn more on the reasoning why you want this feature as I don't understand it fully yet. Can you provide a step by step description on what you are doing? That would be awesome. I am having particular trouble understanding when and how this PR gets created, and what happens when you do a second/third/... mob done.

On a side note: regardless of the feature, one would not to retain the local branch. It's enough to retain the remote branch, as a mob start would automatically (and always) get the remote branch.

hihithisisme:
Hey, thanks for the response! 

> I am having particular trouble understanding when and how this PR gets created, and what happens when you do a second/third/... mob done.
I must acknowledge that I did not think of the implementation details thoroughly enough. This PR will not work upon the second/third... mob done unless we want to force push to the feature branch, which is not ideal. 

Regarding the use case for this feature, perhaps an example would be better. 

1. The team starts on a new mob for feature1. `mob start -b feature1` to create branch `mob/main-feature1`
2. After the entire mobbing process, they do `mob done`
3. Then they push the commit to the `feature1` branch to draft the Pull Request
4. After some comments, the team decides to enter into another mob to iterate on the changes. Now they have to do `mob start` on the feature branch to create branch `mob/feature1`. 

So the pain point here is just that the mob branches for the initial mob and subsequent mobs is different. However, now that I'm actually typing out the workflow, I recognise that perhaps it is better to simply creating the feature branch `feature1` first before initialising the mob branch. 

Sorry for the wild goose chase, but if there's nothing else, I think we can just close this PR. Thanks for the discussion!

simonharrer:
No worries. Thanks for your engagement on improving the mob tool! I love how you proposed your ideas and reflected/participated in the discussion. Kudos to you!

On a side note: with the next mob release, there comes a command named `mob clean`. The more you use feature branches, the more you might want that command.

Closing the PR as requested.

PR #255, opened 2022-04-11
feature: allow user to define a fixed base branch

hihithisisme:
Hi new user here. 
Similar to the https://github.com/remotemobprogramming/mob/pull/254, there's another feature that I would like to suggest. My team typically branches out from the main branch and hence it is redundant to have it in the WIP branch name. It only adds more complexity to the WIP branch name. 
 
I recognise that this might not be accepted since it breaks the original branch naming conventions, but I thought to open it up for discussion anyways. I was able to draft out a proof-of-concept without breaking any of the existing tests. 

Thanks for taking a look!

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,11 +296,20 @@ Mob supports you and can automate this step. You just need the configuration opt
 
 For example if you want use IntelliJ the configuration option would look like this: `MOB_OPEN_COMMAND=idea %s`
 
+### Use a fixed base branch
+
+Say your team typically branches off the same primary branch. It is redundant to state the base branch.
+Setting `MOB_FIXED_BASE_BRANCH` can allow you to remove the base branch name from the wip branch naming convention.
+
+For example, without setting `MOB_FIXED_BASE_BRANCH`, you will have `mob/main-feature1` as the wip branch name.
+Setting `MOB_FIXED_BASE_BRANCH=main` will cause the wip branch to be `mob/feature1` instead.
+
 ## More on Installation
 
 ### Known Issues
 
-- When you have an ssh key with a password and you running mob on windows in powershell, you will not be able to enter a password for your ssh key. You can circumvent this problem by using the git bash instead of powershell.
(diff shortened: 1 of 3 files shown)


simonharrer:
Thanks for reaching out with an already implemented solution. But to be honest, I'm not sure, I want to go in that direction. 

From what I understand is that you want branch names that are more concise. Can you elaborate why longer branch names, or branch names containing redundant info, are an issue for you?

The consequences are that, with this setting, everybody has to have the same value set. Otherwise, they no longer can work together.

In summary, I don't see a real benefit as well as a lot of downsides by adding a flag that MUST be the same for everybody.

hihithisisme:
Hey, thanks for the response! 

It's a good point that this forces everyone to have the same value set. One possible way to enforce this is to only allow this configuration set via the project-specific .mob file. However, I believe your hesitance is more towards whether this is a useful feature to have, so let's focus on that first. 

Use case wise, the longer branch name is definitely not an issue. Just that it adds an additional step in the workflow when switching to or starting a mob from a different branch. Let's take the example of `main` being the base branch and `feature1` being the WIP qualifier. 

Current workflow if you're on a different branch (for example `feature0`):
```
git switch main
mob start -b feature1
```

Proposed workflow (still on different branch `feature0`):
```
mob start -b feature1
```

In short, it would make switching to a mob branch easier (aside from the point on a more concise mob branch name). However, I do see that this might be introducing complexity, so I think it's understandable this feature is not merged.

simonharrer:
Thanks for your detailed explaining of your use case. I, however, like to keep the tool simple. The use case you describe makes the tool more complex, and the benefit is not that huge. That's why I'm closing this PR for now. Thanks for your understanding, despite all your effort on making a first implementation of your proposed feature.

hihithisisme:
Sure. Was a tad hasty in rushing to implementation before determining a proper use case. 
Thanks for taking a look though!

Issue #253, opened 2022-04-07
Allow env var to configure "mob done" squash, but leave manual commits


tedyoung:
Now that `mob done` has three options, the boolean for the env var `MOB_DONE_SQUASH` no longer makes sense. I'd like to default to "squash only wip commits".  Either that env var needs to change, or a new env var is needed, unless I'm missing some other way to configure that default?

gregorriegler:
Thank you, I was looking forward for an issue/pr like this.
When you say you like to default to "squash only wip commits", do you mean that `mob done` by default should do `mob done --squash-wip` ?

tedyoung:
Yes, exactly that.

So either `MOB_DONE_SQUASH=wip` (or `=none` or `=all`), or a new env var `MOB_DONE_SQUASH_WIP=true`.

gregorriegler:
With the current version you may already provide either `squash`, `no-squash` or `squash-wip` as a value for this environment variable, and it should work. However, for backwards compatibility we also still support `true` and `false`. There's now also the new config file which is more strict in it's typing. So we couldn't just have the existing boolean config-file property support strings like the env var does.
We need to decide how to proceed with this.

I would suggest

- we discontinue the support for `true` and `false` in the env var.
- we change the config file var from boolean to string and allow the 3 values.
- we make `squash-wip` the overall default.

@simonharrer what do you think?

simonharrer:
fine for me. let us start with deprecation warnings for true/false values in the next version.

seanpoulter:
Would you like to contribute @tedyoung? If not I'll volunteer to drive this forward.

tedyoung:
My Go is too rusty, @seanpoulter, so would be great for you to drive it.

thmuch:
@seanpoulter I'm working on a PR right now that allows setting `squash-wip` etc. via config file (currently this works only via env var). If you're already working on this in the scope of this issue, please let me know.

seanpoulter:
@tedyoung - I'll give it a try. 🤞 
@thmuch - I look forward to your PR. I'll steer clear of the config file for now.

gregorriegler:
I think we can close this

seanpoulter:
I'm going to reopen this unless you two confirm otherwise.

I thought we still wanted to add a deprecation notice if we set `MOB_DONE_SQUASH` to true or false based on Simon's comment here - https://github.com/remotemobprogramming/mob/issues/253#issuecomment-1094090507.

[...]

PR #256, opened 2022-04-15
Allow setting DoneSquash values via config file

thmuch:
In #253 the feature is mentioned that the three new DoneSquash values can be set via env var right now. This PR allows setting these values via config file, too.

Note: there is a change in behaviour if a garbage value is passed. Before we would issue a warning. Now we match the env var behaviour (no warning, set "squash" value).

Signed-off-by: Thomas Much <github@snailshell.de>


simonharrer:
- missing a test for empty string
- can you add a CHANGELOG entry?

Otherwise LGTM

thmuch:
I'm sorry there's still a bug. New values are read correctly, but assignment to configuration struct has no effect. The unit under test obviously was too small... Working on that.

PR #261, opened 2022-04-26
Add link to fig shell autocompletion

thmuch:
Autocompletion for `mob` is available in the latest stable release from "fig" now.

Is this link enough (IMHO yes)? Or should I supply a screenshot as well?


simonharrer:
Perfect. Screenshot might be too much.

PR #265, opened 2022-05-10
Document short command aliases

thmuch:
I documented the already existing short command aliases in the README and in the man page.

Is this too verbose, too long output? I tried something like

"For some commands there are short aliases available: s (start), n (next), d (done), b (branch), t (timer)"

as a single line below the "Add --debug ..." line. But I think the longer form in this PR is more readable.


simonharrer:
Awesome. Thanks for your help! :-)

PR #274, opened 2022-06-02
Update README.md to match example configuration for MOB_TIMER.

asimpson:
Setting `MOB_TIMER` or having an explicit minutes value, e.g. `mob start 10` is necessary for integration with https://timer.mob.sh to work.


simonharrer:
Thank for your pull request.

The MOB_TIMER value is for having a default value of the timer that is **always** started for every `mob start`. For me, this is an option for power users. And not something that should be enabled by default. The README documents the default values only. 

That's why I am closing this. If you disagree, feel free to reopen this.

asimpson:
@simonharrer I appreciate the feedback and I'm fine with the PR being closed. 

The core problem I was looking to address (and maybe this already exists and I just missed it) was making it clear how the integration with timer.mob.sh works. A timer has to be set either as a param or as configuration for the HTTP request to eventually get sent to timer.mob.sh
https://github.com/remotemobprogramming/mob/blob/c6315dcf31498f6e27571ac776fb67223f951f68/mob.go#L728-L735

Maybe this isn't really an issue but it wasn't clear to me and I spent some time debugging why I wasn't showing up in the room.

simonharrer:
Thanks for sharing more details. How about adding a section talking about the integration with timer.mob.sh in the readme? Would you like to have a first go at it?

asimpson:
@simonharrer Yep yep, I can take a crack at it.

PR #275, opened 2022-06-09
Update go install instruction to v3

Syuparn:
I updated `go install` installation instruction in README to `v3`, which is the latest major version


simonharrer:
Well spotted. Thanks!

Issue #276, opened 2022-06-10
Installing with "--user" on Ubuntu will not put binaries in PATH by default


kejne:
Hi! Love the mob tool! Just installed mob on a team mates Ubuntu PC, but found some issues with the installer. Could do a PR for it, just wanted to discuss it first :slightly_smiling_face: 

When installing using
```
curl -sL install.mob.sh | sh -s - --user
```
in Ubuntu without any specific config, the installer looks like mob install failed.

This seems to be likely due to that the binaries are put in `~/.local/bin`which by default are not part of the path.
Verified that the binaries where indeed put in the dir, but since the mob command won't find the binary, the install indicates an error.

Output looks like:
```
installing latest 'mob' release from GitHub to /home/kejne/.local/bin...
Mob binary location: 
sh: 126: mob: not found
Mob binary version: 

Couldn't find a 'say' command on your system.
While 'mob' will still work, you won't get any spoken indication that your time is up.
Please refer to the documentation how to setup text to speech on a *NIX system.

https://mob.sh#linux-timer

(!) The installation location doesn't match the location of the mob binary.
    This means that the binary that's used is not the binary that has just been installed
    You probably want to delete the binary at 
```

An idea is to add the path to the users `.profile` if the installer detects that the binary is in place, but the command not available.
Thoughts?

simonharrer:
.local/bin will be added to the PATH automatically after a restart after the folder has been created. That’s the issue here. There are several solutions. Perhaps we should give instructions for the user to restart? What do you think?

kejne:
Ah, I did not know that!
Then I think we should be able to alter the install script to make this more clear, since the install looks like a failure today.
I can look at a PR for it. 👌

gregorriegler:
Logout/Login should solve also achieve what the restart does.
I think restart X would also work.

kejne:
I actually just tried to reproduce it, and noticed that the PATH was not updated automatically after reboot/logout for me.
I guess the changes would be the same, just message written that needs to reflect the case.

simonharrer:
Hm, I thought it that way. Perhaps I am wrong. Can someone verify that restart behavior regarding .local/bin added to the classpath after a restart on ubuntu?

kejne:
I could add that I tried with Kubuntu 22.04 and am using zsh. I hadn't installed anything else to ~/.local/bin prior.
After restart I could still not run `mob`, though the executable was put correctly in ~/.local/bin.

[...]

PR #277, opened 2022-06-12
Clarifying error when mob isn't found in PATH

kejne:
Relates to #276 

Gave some hints to the user about what could be the issue and hid the false error message about duplicate binary.
Minor formatting about doc reference as it was not clear that the link was related to the above text.


simonharrer:
Thanks for your contribution! :-)

PR #278, opened 2022-06-13
Test behaviour on pre existing commits

gregorriegler:
Found and addressed a bug where when one of the mobbers has had an unpushed commit thus other mobbers were behind on the base branch. They ran into a problem when they would do `mob done --squash-wip`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 3.1.5
+- Fix: `mob done --squash-wip` now successfully auto-merges auto-mergeable diverging changes.
+
 # 3.1.4
 - Fixes a bug where mob saves the wrong filepath of the last modfied file in the wip commit message.
 
--- a/mob.go
+++ b/mob.go
@@ -1425,7 +1425,7 @@ func done(configuration Configuration) {
 
 		git("checkout", baseBranch.Name)
 		git("merge", baseBranch.remote(configuration).Name, "--ff-only")
-		mergeFailed := gitignorefailure("merge", squashOrNoCommit(configuration), "--ff", wipBranch.Name)
+		mergeFailed := gitignorefailure("merge", squashOrCommit(configuration), "--ff", wipBranch.Name)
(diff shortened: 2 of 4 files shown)


simonharrer:
Which of the new tests would fail with the old behavior?

simonharrer:
Fine for me. Can you add a changelog, please?

gregorriegler:
The more I think about it the more I dislike it.
The wip branch shouldn't diverge from the main branch ever. ( I cannot outrule a conflict )
When it diverges, mobbers should be warned and guided to resolve it instead.

gregorriegler:
@simonharrer what should happen in your opinion if the base branch diverges from the mob branch?

simonharrer:
With squash, it’s not that large of an issue as you normally don’t have conflicts - at least in my experience. And if you do, you have to fix them manually with the staged changes until you make the final commit.

With squash-wip it should behave the same.

With no-squash, it’s more of a problem as it creates a merge commit. But also ok.

simonharrer:
We could add an info to the user that the mob branch diverged from when it was created.

gregorriegler:
I prefer that. If users get the info that it diverges, they at least have the choice to fix it on the wip-branch rather than on the main-branch

gregorriegler:
Tried to rebase my branch onto latest main with IntelliJ which failed and I didn't notice it. 
It had put my commits before all other commits on my branch (really strange). 
So I just undid this and force pushed my branch to do it properly using the CLI.

Added the changelog.

I did some testing. 
The behavior on a non-auto-mergeagle diverging change is just the same as with a simple `mob done`.
This is a behavior that is already well handled. A bit late maybe (on `done`), but its handled.

So I would keep it this way. 
If we want to improve on the behavior on a non-auto-mergeagle conflict later (e.g. detect it earlier) we should do this on a separate branch IMO.

gregorriegler:
@simonharrer resolved the conflict

PR #279, opened 2022-06-14
Resolve install.sh issue & add shell linter to CI

kejne:
My previous PR had a hard to find issue with a missing space in `install.sh`.
Resolved this issue and added a shell linter as part of the CI pipe.

Also resolved issues the linter complained about + excluding non-relevant files where the linter warned because it was unsure.

Relates to #276

### DIFF
--- a/.github/build
+++ b/.github/build
@@ -1,14 +1,14 @@
 #!/bin/bash
 set -eux
 
-PROJECT_NAME=$(basename $GITHUB_REPOSITORY)
+PROJECT_NAME=$(basename "$GITHUB_REPOSITORY")
 NAME="${PROJECT_NAME}_${VERSION}_${GOOS}_${GOARCH}"
 
 EXT=''
 
-if [ $GOOS == 'windows' ]; then
+if [ "$GOOS" == 'windows' ]; then
   EXT='.exe'
 fi
 
-tar cvfz ${NAME}.tar.gz "${PROJECT_NAME}${EXT}" LICENSE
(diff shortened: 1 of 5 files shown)


simonharrer:
Thanks for making the script better.

What happens when the linter finds something? Will the pipeline break?

kejne:
Here is an example of the run from my forked branch before I resolved the linter issues as an example:
https://github.com/kejne/mob/actions/runs/2497313846

(yes, the Test workflow will fail and you will not be able to merge)

simonharrer:
Can you disable that the build fails because of linting issues? I don’t like that.

Otherwise, it’s fine and can be merged.

kejne:
DId you mean that you don't like that you can disable the build from failing? Or that you don't like that the build fails if there is a linter issue?
If you don't like to have the repo linted, I can raise a separate PR with only the bugfix in the install.sh instead :+1: 

The addition of a linter was only to avoid that these kind of issues arise in later changes, since a linter would effectively block PR:s which contain mundane bash formatting issues from being merged.

You cannot simply disable the linter - but you can alter the severity on which it fails:
https://github.com/marketplace/actions/shell-linter

or you can explicitly ignore lines:
https://github.com/koalaman/shellcheck/wiki/ignore

simonharrer:
> Or that you don't like that the build fails if there is a linter issue?

This.

> If you don't like to have the repo linted, I can raise a separate PR with only the bugfix in the install.sh instead 👍

That would be perfect. ❤️

kejne:
Alright, reverted the additional updates, just the fix :+1:

PR #283, opened 2022-06-26
Show help output whenever any kind of help argument is given

SimplyDanny:
As a naïve user (no remembering that the tool has only one help page) I expect to get some help for the `start` command (options that can be specified) when executing `mob start --help`. Instead, a mob session started requiring some manual clean-up from my part afterwards.

I would like to change the behavior in the way that whenever there is some help option in the command, the tool prints the help page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 3.1.5
+- Print the help output whenever any kind of help argument (`help`, `--help`, `-h`) is present in the command, e.g. `mob s 10 -h`.
+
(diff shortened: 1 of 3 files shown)


simonharrer:
Great idea. Can you add a changelog for the next patch version in the `changelog.md`? Otherwise, this looks good to me.

SimplyDanny:
> Can you add a changelog for the next patch version in the `changelog.md`? Otherwise, this looks good to me.

Sure. Done.

PR #287, opened 2022-06-27
Use WIP branch separator in help output

SimplyDanny:
This is a follow-up of dae1648cc46fd which missed this location.


simonharrer:
Hm, tests fail. Can you have a look?

simonharrer:
Rerunning now

Issue #288, opened 2022-06-28
Better error message if git is not installed


simonharrer:
User doesn’t have git installed but installs mob. The error message is not very specific.

PR #289, opened 2022-06-28
ISSUE-288: Update 'git' not installed error message

maciej-stanuch:
Hi, I've decided to help a little with https://github.com/remotemobprogramming/mob/issues/288 
I'm not sure if that's what you meant but feel free to correct me

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # 3.1.5
+- Add a more specific error message if `git` is not installed.
 - Fix: `mob done --squash-wip` now successfully auto-merges auto-mergeable diverging changes.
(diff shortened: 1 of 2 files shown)


simonharrer:
Thanks for the contribution. It goes in the right direction. But instead of parsing the output which is platform and language dependent, it might suffice to only check the status code of the command. Especially, when we use something like git —version, which is independent of whether we are in a git repository. That’s why we need two methods: one check that git is there and one check that revparse works.

maciej-stanuch:
> Thanks for the contribution. It goes in the right direction. But instead of parsing the output which is platform and language dependent, it might suffice to only check the status code of the command. Especially, when we use something like git —version, which is independent of whether we are in a git repository. That’s why we need two methods: one check that git is there and one check that revparse works.

That's a fair point, I will change it accordingly, thanks!

simonharrer:
LGTM. Can you add a CHANGELOG.md entry?

PR #294, opened 2022-07-05
Print a deprecation message when MOB_DONE_SQUASH is set to true or false

### LINKED ISSUE #253: Allow env var to configure "mob done" squash, but leave manual commits

seanpoulter:
This PR:

* should be enough to resolve #253
* displays a deprecation message when `MOB_DONE_SQUASH` is set to `true` or `false`, or the quoted values using an  environment variable or the configuration files
* updates the examples to use values that are not deprecated

--

### Screenshots / Demo

When using environment variables:

Screen Shot 2022-07-05 at 14 49 25

When using the `.mob` configuration file:

Screen Shot 2022-07-05 at 14 51 25


simonharrer:
Awesome! Can you add your changelog to 3.1.5?

seanpoulter:
> Awesome! Can you add your changelog to 3.1.5?

Done. Sorry about that.

thmuch:
Thanks! I like the error handling in `printDeprecatedDoneSquashMessage` and the same layer of abstraction in the calling methods by issuing the warning in a separate method. Thinking about applying both in `setMobDoneSquash` as well (in a different PR).

simonharrer:
> What we don't have yet | no more, though, is a warning for invalid values (like "abc"), do we? Should we issue a warning in that case?

we should. but as part of a new PR.

Issue #293, opened 2022-07-04
mob timer should work anywhere on the system, not just in a git repository


simonharrer:

maciej-stanuch:
Hi @simonharrer, could I be assigned to this issue? 
I would like to try to fix that c:

maciej-stanuch:
PR: https://github.com/remotemobprogramming/mob/pull/295

PR #295, opened 2022-07-06
[#293] Mob timer should work anywhere on the system

maciej-stanuch:
I've added a check for git availability in `getMobTimerRoom` method to allow it to work anywhere in the system.

So, now the behaviour is like this:
- If git repository -> works as it worked before
- If not a git repository -> use the room name from `MOB_TIMER_ROOM` 

---
I wonder if the code, especially `mob.go` file, shouldn't be somehow split into smaller files? It's slightly difficult to work with.
I have not much experience in Go and maybe that's not a good idea to separate all that stuff? I feel like it would make it easier to test it at least.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 3.1.5
 - Add a more specific error message if `git` is not installed.
+- Allow for using `mob timer` outside of git repositories.
 - Fix: `mob done --squash-wip` now successfully auto-merges auto-mergeable diverging changes.
 - Print the help output whenever any kind of help argument (`help`, `--help`, `-h`) is present in the command, e.g. `mob s 10 -h`.
 - Various fixes in suggestion of next typist:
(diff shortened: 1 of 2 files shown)


simonharrer (changes requested):

simonharrer:
We could even write a test for this. As part of the testbed, there's a directory that is not a git repo. We could switch to that and call the main method with `mob timer 0` and verify that we get a specific output in the console. We should change the configuration of the notify command so that no sound output will be generated when running this test.

maciej-stanuch:
> We could even write a test for this. As part of the testbed, there's a directory that is not a git repo. We could switch to that and call the main method with mob timer 0 and verify that we get a specific output in the console. We should change the configuration of the notify command so that no sound output will be generated when running this test.

Will do, I haven't much free time recently to finish it up properly

simonharrer:
Feel free to create a new PR with a test for this.

PR #299, opened 2022-08-15
Add a new command line argument for for disabling SSL verification

JanMeier1337:
It may happen that the SSL verification fails when the mob user has a bad setup in combination with an stupid enterprise network which does SSL interception.

To circumvent that I added a new CLI parameter which makes it possible to disable the SSL verification.

This CLI parameter should only be used when the user is sure what he/she is doing since it could be a security risk to not to verify the certificate.

See the screenshot of how it works:

Bildschirmfoto 2022-08-15 um 09 53 54

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -697,6 +697,7 @@ func config(c Configuration) {
 	say("MOB_TIMER_LOCAL" + "=" + strconv.FormatBool(c.TimerLocal))
 	say("MOB_TIMER_USER" + "=" + quote(c.TimerUser))
 	say("MOB_TIMER_URL" + "=" + quote(c.TimerUrl))
+	say("MOB_TIMER_INSECURE" + "=" + strconv.FormatBool(c.TimerInsecure))
 }
 
 func quote(value string) string {
@@ -1092,8 +1093,8 @@ func sendRequest(requestBody []byte, requestMethod string, requestUrl string, di
 		switch e.Err.(type) {
 		case x509.UnknownAuthorityError:
 			sayError("The timer.mob.sh SSL certificate is signed by an unknown authority!")
-			sayFix("HINT: You can ignore that by adding MOB_TIMER_INSECURE=true to your configuration or environment. Or add is command line parameter:",
-				"mob <your command> --timer-insecure")
+			sayFix("HINT: You can ignore that by adding MOB_TIMER_INSECURE=true to your configuration or environment.",
+				"echo MOB_TIMER_INSECURE=true >> ~/.mob")
 			return fmt.Errorf("failed, to amke the http request: %w", responseErr)
 
 		default:
--- a/mob.go
+++ b/mob.go
@@ -1072,7 +1072,7 @@ func sendRequest(requestBody []byte, requestMethod string, requestUrl string, di
 	sayInfo(requestMethod + " " + requestUrl + " " + string(requestBody))
 
 	responseBody := bytes.NewBuffer(requestBody)
-	request, requestCreationError := http.NewRequest(requestMethod, requestUrl, responseBody)
+	request, requestCreationError := http.NewRequest(requestMethod, "https://untrusted-root.badssl.com/", responseBody)
 
 	httpClient := http.DefaultClient
 	if disableSSLVerification {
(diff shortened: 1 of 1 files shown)


simonharrer:
A few things:

1. The help for this option should only show when you encounter a certificate issue. Can you catch the error and issue the help warning appropriately? I don't want that option in the general help text.
2. The option is one for disabling ssl verification for interaction with the timer.mob.sh only, not in general for the tool. The naming of the option should reflect that. Perhaps use something like `--timer-disable-ssl-verification`
3. Add an option for the configuration files named `MOB_TIMER_DISABLE_SSL_VERIFICATION` that defaults to false.

Perhaps we should also add a section to the README with infos for our enterprise users. We could add this option as well as disabling `no-verify`.

simonharrer:
Even better, we can reuse the parameter naming from curl and name the parameter `--timer-insecure` and `MOB_TIMER_INSECURE`.

JanMeier1337:
Hi @simonharrer, thank you for the feedback.

I am already working on the changes and thinking of to skip the CLI option and make it only available through configuration file: MOB_TIMER_INSECURE=true

What do you think, or should I add the cli parameter too?

mob.go
```diff
@@ -414,6 +418,8 @@ func parseUserConfiguration(configuration Configuration, path string) Configurat
 			setUnquotedString(&configuration.TimerUrl, key, value)
 		case "MOB_STASH_NAME":
 			setUnquotedString(&configuration.StashName, key, value)
+		case "MOB_TIMER_INSECURE":
+			setBoolean(&configuration.TimerDisableSSLVerification, key, value)
```
simonharrer, line 422: Don't forget to parse the variable from project settings as well.

mob.go
```diff
@@ -69,6 +72,7 @@ type Configuration struct {
 	TimerRoomUseWipBranchQualifier bool   // override with MOB_TIMER_ROOM_USE_WIP_BRANCH_QUALIFIER
 	TimerUser                      string // override with MOB_TIMER_USER
 	TimerUrl                       string // override with MOB_TIMER_URL
+	TimerDisableSSLVerification    bool   // override with MOB_TIMER_INSECURE
```
simonharrer, line 75: Let's name the variable `TimerInsecure` for consistency reasons.

mob.go
```diff
@@ -942,7 +949,7 @@ func startTimer(timerInMinutes string, configuration Configuration) {
 
 	if startRemoteTimer {
 		timerUser := getUserForMobTimer(configuration.TimerUser)
-		err := httpPutTimer(timeoutInMinutes, room, timerUser, configuration.TimerUrl)
+		err := httpPutTimer(timeoutInMinutes, room, timerUser, configuration)
```
simonharrer, line 952: I don't want to pass in the configuration to the httpPutTimer method. It's a preparation for extracing this without the configuration to a separate package.

mob.go
```diff
@@ -1043,33 +1050,55 @@ func toMinutes(timerInMinutes string) int {
...
 	}
 
 	request.Header.Set("Content-Type", "application/json")
-	response, responseErr := http.DefaultClient.Do(request)
+	response, responseErr := httpClient.Do(request)
+	if e, ok := responseErr.(*url.Error); ok {
+		switch e.Err.(type) {
+		case x509.UnknownAuthorityError:
+			sayError("The timer.mob.sh SSL certificate is signed by an unknown authority!")
+			sayError("HINT: You can ignore that by adding MOB_TIMER_INSECURE=true to your configuration or environment.")
```
simonharrer, line 1093: Have a look at: `sayTodo`.

simonharrer:
For consistency reasons, I think a CLI parameter makes sense as well.

mob.go
```diff
@@ -1092,7 +1092,8 @@ func sendRequest(requestBody []byte, requestMethod string, requestUrl string, di
 		switch e.Err.(type) {
 		case x509.UnknownAuthorityError:
 			sayError("The timer.mob.sh SSL certificate is signed by an unknown authority!")
-			sayError("HINT: You can ignore that by adding MOB_TIMER_INSECURE=true to your configuration or environment.")
+			sayFix("HINT: You can ignore that by adding MOB_TIMER_INSECURE=true to your configuration or environment. Or add is command line parameter:",
+				"mob <your command> --timer-insecure")
```
simonharrer, line 1096: Can you reproduce the actual command that was used? Should either be timer or break, and add the timer value to it as well. I would like to have the fix as a copy'n'paste solution for the user without any placeholders. For that, we might need to pass more parameters, but I take that into account.

JanMeier1337:
Hi @simonharrer , could you take another look? I would prefer to not add an CLI parameter, since the project has no CLI args library like cobra. The other comments should be solved.

simonharrer:
Your call won't work on windows, but this is perhaps the best solution at the moment.

simonharrer:
So let's get this done. Can you add a CHANGELOG entry?

simonharrer:
Ok, I'll add the changelog myself.

Issue #211, opened 2021-11-09
Feature Request: Create new base branch


jason-cannon:
My git repo is setup such that writing directly to main is not allowed. Instead we have to create a new branch, mob start on that, do our work on it, and then submit a PR to merge that branch into main. 

We felt like it would be useful to have a new mob command to handle the creation of that new base branch for us. I actually ended up forking mob tool to add this functionality in myself

https://github.com/jason-cannon/mob

I added a command called mob create that creates a new base branch, switches to it, and then executes mob start on that branch. Mob Create might not be the best name in the world for it, but it did what we needed. 

```
func create(branchName string, configuration Configuration) {
	git("fetch", configuration.RemoteName)
	if (hasLocalBranch(branchName) || hasRemoteBranch(branchName, configuration)) {
		sayError(fmt.Sprintf("Branch by the name of %s already exists", branchName))
	} else {
		git("pull")
		git("checkout", "-b", branchName)
		git("push","-u",configuration.RemoteName, branchName)
		start(configuration)
	}
}
```

I don't know GO but adding this was simple thanks to the solid foundation you already built. 

I think something like this might be useful for others. It's not that many commands to type in, but it's been convenient for us to be able to do the one command and go. I even use this to create branches that I solo work on for the ease of creating branches and the easy squash of commits. 

Simon suggested that the name mob create might be too generic, and I can definitely agree with that.

gregorriegler:
What do you think about a solution, where you first create a new branch `x` from main, then start using this branch `x` as your main branch? So you run `mob start` on `x`. Once you are finished, you `mob done` on `x`. And then you create a PR from `x` to main.

gregorriegler:
So we're just searching for a better name?
How about `mob fork {branchname}` or `mob feature {branchname}` or `mob from {branchname}`

jason-cannon:
More or less. I implemented this functionality on my own fork but didn't take the time to write unit tests or anything or else I'd just offer a pull request of this functionality. Of those three I think I personally like mob feature

simonharrer:
This is similar to https://github.com/remotemobprogramming/mob/issues/185 so perhaps @lamalex might be interested in this, as well.

An alternative would be to use `mob start --feature {branchname}` instead, making it more obvious that a `mob start` is happening here.

lamalex:
[...]

PR #304, opened 2022-09-16
add function mob start --create

hollesse:
add --create functionality in the simplest form as described in #211

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -60,7 +60,7 @@ type Configuration struct {
 	NotifyMessage                  string // override with MOB_NOTIFY_MESSAGE
 	NextStay                       bool   // override with MOB_NEXT_STAY
 	StartIncludeUncommittedChanges bool   // override with MOB_START_INCLUDE_UNCOMMITTED_CHANGES variable
-	StartCreateRemoteBranch        bool   // override with MOB_START_CREATE_REMOTE_BRANCH variable
+	StartCreate                    bool   // override with MOB_START_CREATE variable
 	StashName                      string // override with MOB_STASH_NAME
 	WipBranchQualifier             string // override with MOB_WIP_BRANCH_QUALIFIER
 	WipBranchQualifierSeparator    string // override with MOB_WIP_BRANCH_QUALIFIER_SEPARATOR
@@ -321,7 +321,7 @@ func getDefaultConfiguration() Configuration {
 		NextStay:                       true,
 		RequireCommitMessage:           false,
 		StartIncludeUncommittedChanges: false,
-		StartCreateRemoteBranch:        false,
+		StartCreate:                    false,
 		WipBranchQualifier:             "",
 		WipBranchQualifierSeparator:    "-",
 		DoneSquash:                     Squash,
@@ -396,8 +396,8 @@ func parseUserConfiguration(configuration Configuration, path string) Configurat
 			setBoolean(&configuration.NextStay, key, value)
 		case "MOB_START_INCLUDE_UNCOMMITTED_CHANGES":
 			setBoolean(&configuration.StartIncludeUncommittedChanges, key, value)
-		case "MOB_START_CREATE_REMOTE_BRANCH":
-			setBoolean(&configuration.StartCreateRemoteBranch, key, value)
+		case "MOB_START_CREATE":
+			setBoolean(&configuration.StartCreate, key, value)
 		case "MOB_WIP_BRANCH_QUALIFIER":
 			setUnquotedString(&configuration.WipBranchQualifier, key, value)
 		case "MOB_WIP_BRANCH_QUALIFIER_SEPARATOR":
@@ -477,8 +477,8 @@ func parseProjectConfiguration(configuration Configuration, path string) Configu
 			setBoolean(&configuration.NextStay, key, value)
 		case "MOB_START_INCLUDE_UNCOMMITTED_CHANGES":
 			setBoolean(&configuration.StartIncludeUncommittedChanges, key, value)
-		case "MOB_START_CREATE_REMOTE_BRANCH":
-			setBoolean(&configuration.StartCreateRemoteBranch, key, value)
+		case "MOB_START_CREATE":
+			setBoolean(&configuration.StartCreate, key, value)
 		case "MOB_WIP_BRANCH_QUALIFIER":
 			setUnquotedString(&configuration.WipBranchQualifier, key, value)
 		case "MOB_WIP_BRANCH_QUALIFIER_SEPARATOR":
@@ -579,7 +579,7 @@ func parseEnvironmentVariables(configuration Configuration) Configuration {
 	setBoolFromEnvVariable(&configuration.NextStay, "MOB_NEXT_STAY")
 
 	setBoolFromEnvVariable(&configuration.StartIncludeUncommittedChanges, "MOB_START_INCLUDE_UNCOMMITTED_CHANGES")
-	setBoolFromEnvVariable(&configuration.StartCreateRemoteBranch, "MOB_START_CREATE_REMOTE_BRANCH")
+	setBoolFromEnvVariable(&configuration.StartCreate, "MOB_START_CREATE")
 
 	setDoneSquashFromEnvVariable(&configuration, "MOB_DONE_SQUASH")
 
@@ -742,7 +742,7 @@ func parseArgs(args []string, configuration Configuration) (command string, para
 		case "--squash-wip":
 			newConfiguration.DoneSquash = SquashWip
 		case "--create":
-			newConfiguration.StartCreateRemoteBranch = true
+			newConfiguration.StartCreate = true
 		default:
 			if i == 1 {
 				command = arg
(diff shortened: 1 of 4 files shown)


mob_test.go
```diff
@@ -742,6 +759,95 @@ func TestStartOnUnpushedFeatureBranch(t *testing.T) {
...
+
+func TestStartCreateOnUnpushedFeatureBranchWithUncommitedChanges(t *testing.T) {
+	output, configuration := setup(t)
+	configuration.StartCreateRemoteBranch = true
+	git("checkout", "-b", "feature1")
+	assertOnBranch(t, "feature1")
+	createFile(t, "file.txt", "contentIrrelevant")
+	start(configuration)
+
+	assertOutputContains(t, output, "To start, including uncommitted changes, use")
```
gregorriegler, line 794: Can we also hint the user of the creation of the branch in this case?

mob_test.go
```diff
@@ -742,6 +759,95 @@ func TestStartOnUnpushedFeatureBranch(t *testing.T) {
...
+	configuration.StartCreateRemoteBranch = true
+	checkoutBranchAndCreateRemoteBranch("feature1")
+	assertOnBranch(t, "feature1")
+	createFile(t, "file.txt", "contentIrrelevant")
+	git("add", ".")
+	git("commit", "-m", "commit ahead")
+	git("push")
+	git("reset", "--hard", "HEAD~1")
+	assertOnBranch(t, "feature1")
+	start(configuration)
```
gregorriegler, line 831: since the tests are getting a bit long, one little thing that might help make those tests easier to read is to move `configuration.StartCreateRemoteBranch = true` down to the act part
```
....

configuration.StartCreateRemoteBranch = true
start(configuration)

assert ...
```
so that you know what this start is about. But I'm not entirely sure, since it really is arranging. Just an Idea.

mob_test.go
```diff
@@ -742,6 +759,95 @@ func TestStartOnUnpushedFeatureBranch(t *testing.T) {
...
+	checkoutBranchAndCreateRemoteBranch("feature1")
+	assertOnBranch(t, "feature1")
+	createFile(t, "file.txt", "contentIrrelevant")
+	git("add", ".")
+	git("commit", "-m", "commit ahead")
+	assertOnBranch(t, "feature1")
+	start(configuration)
+
+	assertOnBranch(t, "feature1")
+	assertOutputContains(t, output, "ERROR cannot start; unpushed changes on base branch must be pushed upstream")
```
gregorriegler, line 849: 'unpushed changes on base branch' is sufficient in my opinion. How the user wants to solve this problem is their decision.

mob.go
```diff
@@ -1155,19 +1164,31 @@ func start(configuration Configuration) error {
...
 	}
 
 	git("fetch", configuration.RemoteName, "--prune")
 	currentBaseBranch, currentWipBranch := determineBranches(gitCurrentBranch(), gitBranches(), configuration)
 
-	if !currentBaseBranch.hasRemoteBranch(configuration) {
+	if !currentBaseBranch.hasRemoteBranch(configuration) && !configuration.StartCreateRemoteBranch {
 		sayError("Remote branch " + currentBaseBranch.remote(configuration).String() + " is missing")
-		sayFix("To set the upstream branch, use", "git push "+configuration.RemoteName+" "+currentBaseBranch.String()+" --set-upstream")
+		sayFix("To start and set the upstream use", "mob start --create")
```
gregorriegler, line 1180: We should tell them this will create a remote branch

mob.go
```diff
@@ -1155,19 +1164,31 @@ func start(configuration Configuration) error {
...
 		return errors.New("remote branch is missing")
 	}
 
+	if !currentBaseBranch.hasRemoteBranch(configuration) && configuration.StartCreateRemoteBranch {
```
gregorriegler, line 1184: could move this block into its own function possibly doing the creation

gregorriegler (changes requested):
Can you please add doc (readme and cli help), as well as a changelog.

mob_test.go
```diff
@@ -42,6 +42,17 @@ func TestParseArgs(t *testing.T) {
 	equals(t, "green", configuration.WipBranchQualifier)
 }
 
+func TestParseArgsStartCreate(t *testing.T) {
+	configuration := getDefaultConfiguration()
+	equals(t, configuration.WipBranchQualifier, "")
```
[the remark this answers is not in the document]
hollesse: oh I forgot to delete that.

mob_test.go
```diff
@@ -42,6 +42,17 @@ func TestParseArgs(t *testing.T) {
...
+
+	command, parameters, configuration := parseArgs([]string{"mob", "start", "--create"}, configuration)
+
+	equals(t, "start", command)
+	equals(t, "", strings.Join(parameters, ""))
+	equals(t, true, configuration.StartCreateRemoteBranch)
```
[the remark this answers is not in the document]
hollesse: good point :)

mob_test.go
```diff
@@ -555,8 +573,7 @@ func TestStartWarnsOnDivergingWipBranch(t *testing.T) {
 func TestStartNextOnFeatureWithBranch(t *testing.T) {
 	_, configuration := setup(t)
 	configuration.WipBranchQualifier = "green"
-	git("checkout", "-b", "feature1")
-	git("push", "origin", "feature1", "--set-upstream")
+	checkoutBranchAndCreateRemoteBranch("feature1")
```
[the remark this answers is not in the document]
hollesse: `checkoutAndPushBranch` sounds good :)

mob_test.go
```diff
@@ -742,6 +759,95 @@ func TestStartOnUnpushedFeatureBranch(t *testing.T) {
 
 	assertOnBranch(t, "feature1")
 	assertOutputContains(t, output, "Remote branch origin/feature1 is missing")
+	assertOutputContains(t, output, "mob start --create")
+}
+
+func TestStartOnUnpushedFeatureBranchWithUncommitedChanges(t *testing.T) {
+	output, configuration := setup(t)
+	git("checkout", "-b", "feature1")
+	assertOnBranch(t, "feature1")
```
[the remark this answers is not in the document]
hollesse: I agree

mob_test.go
```diff
@@ -742,6 +759,95 @@ func TestStartOnUnpushedFeatureBranch(t *testing.T) {
...
+	assertOutputContains(t, output, "mob start --include-uncommitted-changes")
+}
+
+func TestStartCreateOnUnpushedFeatureBranch(t *testing.T) {
+	_, configuration := setup(t)
+	configuration.StartCreateRemoteBranch = true
+	git("checkout", "-b", "feature1")
+	assertOnBranch(t, "feature1")
+	start(configuration)
```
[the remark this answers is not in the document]
hollesse: I agree. I did not even notice that the others tests used this pattern.

mob_test.go
hollesse: I agree :)

mob_test.go
hollesse: I could change it, but thats the behavior it hab before ;)

mob.go
hollesse: isn't it enough to say that it sets the upstream? Isn't that the same than creating a remote branch?

mob_test.go
hollesse: done.

mob.go
gregorriegler: The way I understand it is that you can create a local branch "a", and you can then create a remote branch "origin/b", and they are not connected in any way. Even the names are different, "a" vs "b". Then you can choose to tell your local git to connect your local branch "a" to the remote branch "origin/b", by saying "--set-upstream".
So I'm thinking set-upstream and create a remote branch are two different things, and set-upstream is just doing the connection between the two in your local git.
But maybe you are right, and we don't yet create the remote branch. This will happen only later as we push?

gregorriegler (commented):
thank you for being patient with me.
I think since this adds a new feature, we might want to increase the minor number instead.
so 3.3.0.

mob.go
```diff
@@ -1211,6 +1226,16 @@ func start(configuration Configuration) error {
...
 }
 
+func createRemoteBranch(configuration Configuration, currentBaseBranch Branch) {
+	if !currentBaseBranch.hasRemoteBranch(configuration) && configuration.StartCreate {
+		git("push", configuration.RemoteName, currentBaseBranch.String(), "--set-upstream")
+	}
+
+	if currentBaseBranch.hasRemoteBranch(configuration) && configuration.StartCreate {
+		debugInfo("Remote branch " + currentBaseBranch.remote(configuration).String() + " already exists")
+	}
```
simonharrer, line 1236: The first if creates the remote branch and the second if then detect that newly created remote branch.  Is that expected behavior?

gregorriegler: Good catch, I think this needs an else. Also we could consider making the debugInfo an info

mob.go
hollesse: okay I see your point. I knew that this is possible but i never saw a usecase for that, because of this for me it felt similar. But then I will rename it and call it "create a remote branch". No the remote branch is directly created and pushed so i think the naming above is correct.

hollesse:
> thank you for being patient with me. I think since this adds a new feature, we might want to increase the minor number instead. so 3.3.0.

I bumped the version up to 3.3.0

mob.go
hollesse: Very good catch :) added it to the tests and fixed it :)

hollesse:
> Nice one! Can you please add doc (readme and cli help), as well as a changelog.

done.

hollesse:
I will a another test, to test the create parameter together with branch parameter

gregorriegler:
I think I merged early, you still wanted to add that test.
But it's not a problem, you can still add it.

PR #306, opened 2022-09-17
fixes typo and adjusts link to tech radar

sebsprenger:
as the old link was pointing to the current volume and not the one that featured mob


hollesse:
Hi @sebsprenger thank you for this PR

Issue #296, opened 2022-07-07
Scripts getting long


gregorriegler:
We're close to 2000 lines in mob.go and its getting a little painful to work with it.
I can't find what I'm searching for by scrolling. I use find, and I'm always a little lost.
And it's the same with mob_test.go.
I think it would be a good idea to start and extract cohesive functionality to their own files.
I'm thinking about things like:

- configuration (loading it from cli, files and the struct)
- branch
- timer
- and potentially others too

Would love to hear your opinions on this.

simonharrer:
At the moment, everything is highly coupled. That's why having it in one file is better than having it in more files but still being highly coupled (at least from my point of view). The current "architecture" already contains circular dependencies, which is really bad... I documented the current architecture in the docs folder as a plantuml file. Here's the current rendering:

http://www.plantuml.com/plantuml/png/SoWkIImgAStDuIf8JCvEJ4zLo4tCp8DpqZBpKkBpKYjAu2B9-Ocb9HdvHSLmaUB2qiJYZFZop08kBf1c1LrTEuH6Y92DRavgMWbG7AuT9cw7hDyXDIy5w1O0

The code is growing with more features, so we need to do something at some point. We need to decouple when we want to have multiple files. For that, we need clear interfaces between the parts, and they should not have circular dependencies. This will be costly, and there might never be a return of investment. Nevertheless, we can try to go forward.

The key question for me is, how we can achieve a better structure without circular dependencies and is still easy to understand and maintain. And this boils down to the interfaces between the parts, and what parts we need.

We would need some kind of logging/console-output package, a git-cli package, etc.

gregorriegler:
We can focus on one leaf at a time, and try to extract that.
While doing that we will notice whether there are any cyclic dependencies. And if there are any, we would address that. If we can't right now, we focus on another leaf. And then we can try and minimize the interface of the newly extracted module.

I would start with a simple thing.

gregorriegler:
The PR shows how we could extract leaves without introducing cyclic dependencies.

Now we would have 
```
mob --------------------------⬎
    ⤷configuration ------> say
```

simonharrer:
I experimentally refactored a little bit this evening as well.

There are three things that almost every function uses: git, configuration, and say. We can sometimes mitigate git or configuration by passing in the necessary infos instead of getting them within a package, but probably not always does this lead to better code. I am still not sure which direction to take here ... or if this is currently the way to go.

[...]

PR #312, opened 2022-09-20
Refactor: Extract say to its own file

gregorriegler:
A first step tackling #296

### DIFF
--- a/coauthors.go
+++ b/coauthors.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"github.com/remotemobprogramming/mob/v3/say"
 	"os"
 	"path"
 	"regexp"
@@ -24,20 +25,20 @@ func collectCoauthorsFromWipCommits(file *os.File) []Author {
 	// For details and background, see https://github.com/remotemobprogramming/mob/issues/81
 
 	coauthors := parseCoauthors(file)
-	debugInfo("Parsed coauthors")
-	debugInfo(strings.Join(coauthors, ","))
+	say.Debug("Parsed coauthors")
+	say.Debug(strings.Join(coauthors, ","))
 
 	coauthors = removeElementsContaining(coauthors, gitUserEmail())
-	debugInfo("Parsed coauthors without committer")
-	debugInfo(strings.Join(coauthors, ","))
+	say.Debug("Parsed coauthors without committer")
+	say.Debug(strings.Join(coauthors, ","))
 
 	coauthors = removeDuplicateValues(coauthors)
-	debugInfo("Unique coauthors without committer")
-	debugInfo(strings.Join(coauthors, ","))
+	say.Debug("Unique coauthors without committer")
+	say.Debug(strings.Join(coauthors, ","))
 
 	sortByLength(coauthors)
-	debugInfo("Sorted unique coauthors without committer")
-	debugInfo(strings.Join(coauthors, ","))
+	say.Debug("Sorted unique coauthors without committer")
+	say.Debug(strings.Join(coauthors, ","))
 
 	return coauthors
 }
@@ -94,11 +95,11 @@ func removeDuplicateValues(slice []string) []string {
 
 func appendCoauthorsToSquashMsg(gitDir string) error {
 	squashMsgPath := path.Join(gitDir, "SQUASH_MSG")
-	debugInfo("opening " + squashMsgPath)
+	say.Debug("opening " + squashMsgPath)
 	file, err := os.OpenFile(squashMsgPath, os.O_APPEND|os.O_RDWR, 0644)
 	if err != nil {
 		if os.IsNotExist(err) {
-			debugInfo(squashMsgPath + " does not exist")
+			say.Debug(squashMsgPath + " does not exist")
 			// No wip commits, nothing to squash, this isn't really an error
 			return nil
 		}
--- a/mob.go
+++ b/mob.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/remotemobprogramming/mob/v3/say"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -27,7 +28,6 @@ const (
 
 var (
 	workingDir                 = ""
-	Debug                      = false // override with --debug parameter
 	GitPassthroughStderrStdout = false // hack to get git hooks to print to stdout/stderr
 )
 
@@ -129,8 +129,8 @@ func (branch Branch) remote(configuration Configuration) Branch {
 func (branch Branch) hasRemoteBranch(configuration Configuration) bool {
 	remoteBranches := gitRemoteBranches()
 	remoteBranch := branch.remote(configuration).Name
-	debugInfo("Remote Branches: " + strings.Join(remoteBranches, "\n"))
-	debugInfo("Remote Branch: " + remoteBranch)
+	say.Debug("Remote Branches: " + strings.Join(remoteBranches, "\n"))
+	say.Debug("Remote Branch: " + remoteBranch)
 
 	for i := 0; i < len(remoteBranches); i++ {
 		if remoteBranches[i] == remoteBranch {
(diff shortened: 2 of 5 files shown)


hollesse:
I really like the idea, but I would have expected it to be its own package and export just few functions and not all. What do you think about this?

say.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func sayFix(instruction string, command string) {
+	sayWithPrefix(instruction, "👉 ")
+	sayEmptyLine()
+	sayIndented(command)
+	sayEmptyLine()
+}
+
+func sayNext(instruction string, command string) {
```
hollesse, line 47: then this should be public

say.go
```diff
@@ -0,0 +1,80 @@
...
+	sayWithPrefix(instruction, "👉 ")
+	sayEmptyLine()
+	sayIndented(command)
+	sayEmptyLine()
+}
+
+func sayWithPrefix(s string, prefix string) {
```
hollesse, line 54: here I am not sure if this should be public or private

say.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func say(s string) {
+	if len(s) == 0 {
+		return
+	}
+	printToConsole(strings.TrimRight(s, " \r\n\t\v\f") + "\n")
+}
+
+func debugInfo(text string) {
```
hollesse, line 72: then this should be public

say.go
```diff
@@ -0,0 +1,80 @@
...
+	if Debug {
+		sayWithPrefix(text, "DEBUG ")
+	}
+}
+
+var printToConsole = func(message string) {
```
hollesse, line 78: why is this an anonymous function?

hollesse (changes requested):
I think we should use a different package and have some functions private

gregorriegler:
You're just teaching me some go, and I love it!

say.go
gregorriegler: tried making it a func. found out that its overridden in mob_test so leaving it for now.

gregorriegler:
I'm stuck. The tests finish with an exit code and I don't know why.

gregorriegler:
Ok, I understand now. Trying to find a solution

gregorriegler:
solved

coauthors.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"github.com/remotemobprogramming/mob/v3/say"
```
hollesse, line 6: I'am not sure if this is the way to go. But disclaimer I also didn't knew that before I just looked up if there is something like private. For me this looks a bit strange but maybe I do not know how it correctly works.

say/say.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func WithPrefix(s string, prefix string) {
+	lines := strings.Split(strings.TrimSpace(s), "\n")
+	for i := 0; i < len(lines); i++ {
+		PrintToConsole(prefix + strings.TrimSpace(lines[i]) + "\n")
+	}
+}
+
+func EmptyLine() {
```
hollesse, line 65: I think this could be private, so it should be `func emptyLine() {`

say/say.go
```diff
@@ -0,0 +1,84 @@
...
+	PrintToConsole("\n")
+}
+
+func Say(s string) {
```
hollesse, line 69: I saw this is used, but should we really use it outside of this package? I am not sure...

say/say_test.go
```diff
@@ -0,0 +1 @@
+package say
```
hollesse, line 1: Do we need that file if we do not have tests for that?

squash_wip.go
```diff
@@ -69,7 +70,7 @@ func getEnvGitEditor() (gitEditor string, gitSequenceEditor string) {
 func mobExecutable() string {
 	if isTestEnvironment() {
 		wd, _ := os.Getwd()
-		return "go run $(ls -1 " + wd + "/*.go | grep -v _test.go)"
+		return "cd " + wd + " && go run ."
```
hollesse, line 73: wow i did not expect that :D

say/say.go
gregorriegler: Its used a lot it seems. could rename it, but I think we need it.

say/say_test.go
gregorriegler: at first it asked me for it. but i will try and delete it again

coauthors.go
hollesse: I just read a bit about packages. This should be fine :)

say/say.go
hollesse: The name is fine but I was just thinking if this is something more general which should not be used. But for now it's fine to just leave it as it is.

say/say_test.go
hollesse: If we need to keep it we could also just write some tests ;)

gregorriegler:
ok, we got this.

mob.go
```diff
@@ -664,55 +654,55 @@ func printDeprecatedDoneSquashMessage(value string) {
 
 	if unquotedValue == "true" || unquotedValue == "false" {
 		newValue := doneSquash(unquotedValue)
-		say("MOB_DONE_SQUASH is set to the deprecated value " + value + ". Use the value " + newValue + " instead.")
+		say.Say("MOB_DONE_SQUASH is set to the deprecated value " + value + ". Use the value " + newValue + " instead.")
```
hollesse, line 657: Shouldn't this be a Warning?

mob.go
```diff
@@ -664,55 +654,55 @@ func printDeprecatedDoneSquashMessage(value string) {
...
 	}
 }
 
 func removed(key string, message string) {
 	if _, set := os.LookupEnv(key); set {
-		say("Configuration option '" + key + "' is no longer used.")
-		say(message)
+		say.Say("Configuration option '" + key + "' is no longer used.")
```
hollesse, line 663: This can also be a Warning

mob.go
```diff
@@ -664,55 +654,55 @@ func printDeprecatedDoneSquashMessage(value string) {
...
+		say.Say(message)
 	}
 }
 
 func deprecated(key string, message string) {
 	if _, set := os.LookupEnv(key); set {
-		say("Configuration option '" + key + "' is deprecated.")
-		say(message)
+		say.Say("Configuration option '" + key + "' is deprecated.")
+		say.Say(message)
```
hollesse, line 671: same here

mob.go
```diff
@@ -664,55 +654,55 @@ func printDeprecatedDoneSquashMessage(value string) {
...
 	}
 }
 
 func experimental(key string) {
 	if _, set := os.LookupEnv(key); set {
-		say("Configuration option '" + key + "' is experimental. Be prepared that this option will be removed!")
+		say.Say("Configuration option '" + key + "' is experimental. Be prepared that this option will be removed!")
```
hollesse, line 677: same here

mob.go
```diff
@@ -877,10 +867,10 @@ func (branch Branch) isOrphanWipBranch(configuration Configuration) bool {
 }
 
 func branch(configuration Configuration) {
-	say(silentgit("branch", "--list", "--remote", newBranch("*").addWipPrefix(configuration).remote(configuration).Name))
+	say.Say(silentgit("branch", "--list", "--remote", newBranch("*").addWipPrefix(configuration).remote(configuration).Name))
```
hollesse, line 870: does that make sense? why don't we use `git` instead of `say.Say(silentgit(...))`

mob.go
```diff
@@ -877,10 +867,10 @@ func (branch Branch) isOrphanWipBranch(configuration Configuration) bool {
...
 
 	// DEPRECATED
-	say(silentgit("branch", "--list", "--remote", newBranch("mob-session").remote(configuration).Name))
+	say.Say(silentgit("branch", "--list", "--remote", newBranch("mob-session").remote(configuration).Name))
```
hollesse, line 873: same here

hollesse (approved):

gregorriegler:
I agree with all the points you made, however it is not a part of the change I'm trying to make in this pr. I just want to move say to its own package. We should improve how say is used, but that's another step. Let's make small steps and integrate often.

hollesse:
I agree. That's why I approved the PR but leaved the comments.

gregorriegler:
Ah, I didn't get that. So I'll merge it then.

PR #313, opened 2022-09-22
added tests and bump module version

hollesse:


mob_test.go
```diff
@@ -643,6 +643,37 @@ func TestResetDeleteRemoteWipBranchCommit(t *testing.T) {
...
+	reset(configuration)
+
+	assertOutputContains(t, output, "mob reset --delete-remote-wip-branch")
+	assertMobSessionBranches(t, configuration, "mob/master-green")
+}
+
+func TestResetDeleteRemoteWipBranchCommitBranch(t *testing.T) {
+	_, configuration := setup(t)
+	configuration.WipBranchQualifier = "green"
+	configuration.ResetDeleteRemoteWipBranch = true
```
gregorriegler, line 664: We could move this down, to the 'act' part, like we did in the other tests.
But I'm fine with this, too.

mob_test.go
```diff
@@ -643,6 +643,37 @@ func TestResetDeleteRemoteWipBranchCommit(t *testing.T) {
 	assertNoMobSessionBranches(t, configuration, "mob-session")
 }
 
+func TestResetCommitBranch(t *testing.T) {
+	output, configuration := setup(t)
+	configuration.WipBranchQualifier = "green"
+
```
gregorriegler, line 649: I would delete that empty line, but its not a big deal

mob_test.go
```diff
@@ -643,6 +643,37 @@ func TestResetDeleteRemoteWipBranchCommit(t *testing.T) {
...
+	assertOutputContains(t, output, "mob reset --delete-remote-wip-branch")
+	assertMobSessionBranches(t, configuration, "mob/master-green")
+}
+
+func TestResetDeleteRemoteWipBranchCommitBranch(t *testing.T) {
+	_, configuration := setup(t)
+	configuration.WipBranchQualifier = "green"
+	configuration.ResetDeleteRemoteWipBranch = true
+
```
gregorriegler, line 665: I would delete that empty line, but its not a big deal

gregorriegler (approved):

mob_test.go
hollesse: That doesn't work here, as we need this configuration before we do mob start. I can duplicate it to have if in the act part to if you like.

PR #314, opened 2022-09-23
WIP: start to extract git

hollesse:
I started extracting git into it's own package

### DIFF
--- a/coauthors.go
+++ b/coauthors.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"github.com/remotemobprogramming/mob/v4/git"
 	"github.com/remotemobprogramming/mob/v4/say"
 	"os"
 	"path"
@@ -28,7 +29,7 @@ func collectCoauthorsFromWipCommits(file *os.File) []Author {
 	say.Debug("Parsed coauthors")
 	say.Debug(strings.Join(coauthors, ","))
 
-	coauthors = removeElementsContaining(coauthors, gitUserEmail())
+	coauthors = removeElementsContaining(coauthors, git.UserEmail())
 	say.Debug("Parsed coauthors without committer")
 	say.Debug(strings.Join(coauthors, ","))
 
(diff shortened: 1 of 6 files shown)


hollesse:
@gregorriegler one thing I found out is, that it maybe would make sense to have some generic methods to read configuration, but then let each package have it's own configuration. also the git package could ne it's own configuration as we have some configs just for git. Like Git Hooks or Remote name. What do you think about that?

gregorriegler:
@hollesse

I see two options: 

1.)
Each package defines and owns its configuration properties, but does not know where they're coming from. The props might just be a boolean, as in the say package, but they might also be multiple values, defined as a cohesive struct. so if you want to use a module, you give it the properties it wants, and bam it's ready to use. 

In this case, there needs to be some other role that is responsible for giving the props to the modules, and necessarily initialize them. I like to call it `application`. From a dependency perspective, this `application` sits at the top of the tree and nothing depends on it. The `application` initializes everything and wires it up, or lets say, composes the whole system. 

But where does the `application` get all the configuration properties from? So then there needs to be yet another role responsible for reading props from the given sources.  

2.)
A module completely configures and initializes itself. So then it needs a way to get the properties from our configuration sources. How? They could all just do it themselves, but it would lead to a lot of duplication. So we could have a shared configuration package they could all use to grab certain props from our configuration sources.

---

The challenge in each of those is: We don't want circular dependencies. So whenever we extract a module, we need to make sure it only depends on everything else, or it is a leaf (everything else depends on it)

hollesse:
@gregorriegler so the approach I was thinking about is your second suggestion with the shared code package for how to read configs in a generic way. I think I like this approach more as the application or main package does not need to now about the configuration and the commands and packages themself would be more self contained. But on the other hand I think this would need much more refactoring. Maybe we should go for the first approach. Because it is easy to split right now and enables us to extract the commands. After that is done we could have a look again and refactor it like you described in the second approach.

gregorriegler:
fine with me

hollesse:
I've got a feeling that this PR was the wrong approach. I will start extracting help command and everything which is needed for this and then extracting the reset command. I will close this PR.

PR #316, opened 2022-09-26
extracted help command

hollesse:


gregorriegler (commented):
Such a big PR, phew 😅.

I'm all for extracting the configuration. It takes so much noise out of `mob.go`, so I think it's a crucial next step. However there are a couple things I don't fully understand.

1. Why did you split configuration.go and read_configuration.go? What is the reasoning behind that decision in terms of design?
2. You removed some of the configuration methods that only used the configurations values. What is the reasoning behind that decision?
3. You also kept the `Mob` method on the configuration. What is the reason behind keeping exactly that command?

hollesse:
Yes it is a big PR but it was surprisingly straight forward and was done in under an hour of work.

1) I have splitten configuration and read configuration because It gave me a better overview which of the functions are methods of the configuration. At this point it is not much, but I had difficulties with the branch struct and it's methods. I wanted to try here how it feels to have it in a separate file. I like it. What do you think about it? 
2) I changed the methods to functions because I would like to have the configuration as small as possible in the beginning. If we find functions that are used by multiple packages in the progress of extracting the other commands I would make them a method again and add it to the package, but I don't want to add all right now, as I don't wanted to look where it is used. I just wanted to stay focused on the help command and everything it needed. 
3) I really don't know why I left it there. Right now I believe this is a crucial part. But we could also make it a function for now. And refactor it with the other functions when extracting the other commands.
4) I really like selfdescribing names. Config is well know as abbreviation for configuration but it is a abbreviation and could lead to more abbreviations, this is the main reason why I used the Alias instead of renaming the package. For now I also had no a better name in mind.

gregorriegler:
This is all about Modularity, Coupling and Cohesion.

Ad 1. In my opinion its important to keep the behavior close to the data it operates on, to form cohesive units. I see many advantages with this approach. When I see a method that uses some data, and I see the definition of the data on the same screen, that helps me a lot. However, I don't see any advantage in splitting the data from the behavior like this. One might argue that it is about splitting by the 'kind' of code. However, I don't see any rationale in that. I think its made up.
One argument would be, that the files will be smaller. I see that, but in that case I would aim to split the data, and have multiple files contain both data and behavior.
Ad 2. That's again playing in the cohesion space. If there is a function that only uses data of X. It is more cohesive to have the function near X. If it's not a method at X, there's a risk that we forget that the problem is already solved in another file/package, and we implement it again. This creates a hazard where duplication strives. When it is already a method on X, we won't create duplication, as we will use whats already there. Also it is about information hiding. If configuration knows how a wip branch qualifier is setup, its consumers don't have to. I'm not saying configuration is the right place to do that. But right now I don't see a different place. Information Hiding is good, because it decouples.
Ad 3. Until we have a better place where those functions can live, as long as a function only uses data of configuration, such as that it derives the configuration data, I think we should leave it at configuration as a method. As a result a consumer does not need to know or consume as much of configuration.
Ad 4. I am with you. I also dislike abbreviations. Config is a middle ground. It's a commonly used name. I also don't like, that we have config and configuration. There is no way to distinguish.

hollesse:
1) My problem with taking it together is how to define methods in go. As we do not have a Class which holds all the methods together the methods could be distributed over the home configuration file. And then it's hard to see what methods configuration has. A compromise would be to have it in one file but the methods should all be directly under the struct declaration.

2) and 3) I see your point. I agree that all functions which only use configuration data should be in the configuration package. But if it takes any other argument I would leave it in mob.go for the first step and maybe extract it later if it is used in different commands. I also don't want to make the configuration to big in the initial step such that we can integrate faster. 

4) It is really bad that we use config and configuration but we can not call both configuration as Go the does not now if we mean our variable or our package. We should find a better naming for one of them. I thought about naming the configuration variable configurationFile but that is not correct in all the cases as it could also just be the default configuration or derived from Environment variables.

gregorriegler:
We had some methods, such as this:
```
func (c Configuration) wipBranchQualifierSuffix() string {
	return c.WipBranchQualifierSeparator + c.WipBranchQualifier
}
```
Since it only uses configuration, I think it makes sense to keep it as a method and have it below the struct. I'm not saying it's the best place. Maybe a branch package would be a better place in the future. But for now, I see no reason to split most of those methods from the struct. I would wait with this decision until we have better evidence.

Let's just leave config like you did it. We can always change it when we find a better way.

hollesse:
Okay. I will add the methods again an merge both files together. I will hopefully finish it today. Do you have anything else you would have done differently?

simonharrer:
LGTM

PR #323, opened 2022-10-03
Removes the deprecated configuration property: MOB_START_INCLUDE_UNCO…

gregorriegler:
…MMITTED_CHANGES


configuration/configuration.go
```diff
@@ -535,7 +516,7 @@ func experimental(key string) {
 
 func doneSquash(value string) string {
```
hollesse, line 517: Is this function need anymore? If someone sets the value to a value that does not exist this should be handled within mob done squash or isn't it that way?

gregorriegler: Tried removing it. But it is what handles turning "garbage" into a default. Could code that differently, but I did not come up with a simpler solution, so I left it.

hollesse: Ok fine for me

PR #325, opened 2022-10-04
simplified findMobTimerProcessIdsLinuxAndDarwin and fix mac CI build

hollesse:


gregorriegler:
LGTM. The mac build fails, also on the main branch. Not related to this change however.

hollesse:
@gregorriegler the mac build is working now :)

mob_test.go
```diff
@@ -1508,13 +1507,14 @@ func TestHelpRequested(t *testing.T) {
 }
 
 func TestAbortTimerIfNewTimerIsStarted(t *testing.T) {
-	_, configuration := setup(t)
+	output, configuration := setup(t)
+	say.TurnOnDebugging()
```
gregorriegler, line 1511: wir haben hier noch debugging an

mob_test.go
```diff
@@ -1528,35 +1528,41 @@ func assertNoTimerProcess(t *testing.T) {
 }
 
 func TestAbortBreakTimerIfNewBreakTimerIsStarted(t *testing.T) {
-	_, configuration := setup(t)
+	output, configuration := setup(t)
+	say.TurnOnDebugging()
```
gregorriegler, line 1532: auch hier

mob_test.go
```diff
@@ -1528,35 +1528,41 @@ func assertNoTimerProcess(t *testing.T) {
...
 
+	assertOutputContains(t, output, "Killed mob timer with PID")
 	assertSingleTimerProcess(t)
 	abortRunningTimers()
 }
 
 func TestAbortTimerIfMobNext(t *testing.T) {
-	_, configuration := setup(t)
+	output, configuration := setup(t)
+	say.TurnOnDebugging()
```
gregorriegler, line 1545: debugging

mob_test.go
```diff
@@ -1528,35 +1528,41 @@ func assertNoTimerProcess(t *testing.T) {
...
 	next(configuration)
 
+	assertOutputContains(t, output, "Killed mob timer with PID")
 	assertNoTimerProcess(t)
 }
 
 func TestAbortTimerIfMobDone(t *testing.T) {
-	_, configuration := setup(t)
+	output, configuration := setup(t)
+	say.TurnOnDebugging()
```
gregorriegler, line 1558: debugging

gregorriegler (approved):

gregorriegler:
ok, now I understand that it relies on the debugger. Thats weird. Let me think about how we can fix that.

hollesse:
Correct at the moment we just have debug logs for it. Maybe it makes sense to have a info log that tells the user that the timer was overwritten or aborted. But I'm not sure if the user really needs that info or if this would be noise as this is behavior the new user would expect anyways

gregorriegler:
I made the change to just not check that. We know that the Process is there, so we know it must have been removed. No need to check the logs. And now we don't change the log level for other tests. If we still want to test whether it was invoked, this should in my opinion be done on a more granular level, where the kill function is injected and replaced with a spy.
If the test are green, I'm fine with merging this.

hollesse:
Ah correct I missed that this is the activated for all tests. Then I am fine with removing this assert. Just merge it if the tests are green

PR #327, opened 2022-10-05
rewrite testbed to go

hollesse:
I really just did a rewrite. I would like to integrate it and afterwards refactor it that we do not clone the repo multiple times if not needed for the test.


gregorriegler:
The build is still failing...

hollesse:
This is really strange. This commit should not fix this, but I don’t know why it is failing. I worked before and we didn’t change the code but after merging to master it did not work. And on my Mac I can verify that it works… that’s really strange

gregorriegler:
Do you use go 1.16.1 in your mac?

gregorriegler:
I'm a bit lost how we continue with this. I think we should focus on the build first, but my problem is that I don't have a Mac System. 
It seems to be failing because of the logic that handles the timer processes on mac.
And I don't quite understand why it would take 7 minutes to fail.
Simplest way would be to reproduce it on a local machine for fast feedback, and fix it from there.
Developing in the pipeline is also possible, but quite slow and tedious.
As long as this is failing, I'm not confident that the timer feature works on every mac setup.
I'm also open to pair on this issue.
What do you think @hollesse ?

hollesse:
I am missing more and more confidence in GitHub actions with macOS. Today I did not had any time to check it, but for me this issue felt more like it failed without any reasonable changes. If you want to and have time we could pair tomorrow. I planned to work every Friday afternoon on the mob tool. I will start tomorrow after lunch at 1pm. Just let me know and we can meet in Zoom.

gregorriegler:
This weekend is not easy for me. Maybe Saturday evening. But I'm also open to turn off the mac build if you tested it on your mac manually and found it works just fine.

mob_test.go
```diff
@@ -1759,3 +1767,130 @@ func checkoutAndPushBranch(branch string) {
...
+	git("symbolic-ref", "HEAD", "refs/heads/"+branch)
+	say.Info("finished")
+}
+
+func cloneRepository(path, remoteDirectory string) {
+	say.Info("clonerepo: Cloning remote " + remoteDirectory + " to " + path)
+	err := os.MkdirAll(path, 0755)
+	if err != nil {
+		say.Error("Could not create directory " + path)
+		say.Error(err.Error())
```
gregorriegler, line 1797: interesting to me that we do not return here

gregorriegler:
Made minor rewordings, LGTM

hollesse:
I would like to change some things here. First I would like to return after error as you discovered this is missing. 

I would also like to print all errors directly with fmt. Such that they do not appear in the captured output. And that errors are directly visible. 

I also thought about replacing the say.Info either with Println directly or with say.Debug such that these logs do not appear in the output, as these logs are not part of the Programm. Which version do you prefer @gregorriegler

gregorriegler:
I'd vote for say.Debug if the function name is included in the message, like it is now. Because to me it looks like a developers info, as it includes what function has been called. say.Info would also work, but then it should not include the function name, but just what the outcome is. I would avoid printing and channel everything through say.

hollesse:
But does it make sense to use say here, as these are test logs and nothing which the user can see in the application? I would like to handle that differently as we collect all logs via say and assert against it if a user could have seen a specific message

gregorriegler:
Ok, I didn't think about it this way. I thought about say as a sort of a logging facade, because it has those functions 'Info' 'Debug' etc. But you're right. Say is more an Output to a user than a log. Maybe creating say.Debug was a mistake to begin with. I'm just a bit careful about using print directly. If you use print it becomes sort of a hardcoded channel to stdout, that is not easily reconfigurable.

hollesse:
I don't see say.Debug as a problem. As the user also wants to see debug infos, but not always. But I see a problem to use it within tests. But it could also be ok to not assert debug logs, as this shouldn't be something the user is always interested in. I'am also thinking about other solutions but I do not have an idea right now.

gregorriegler:
Ok, so we're not sure, then lets do what is simplest. We can change it later when we are more confident.

PR #330, opened 2022-10-12
ci with custom git version

hollesse:
I adjusted the CI pipeline, such that it can be tested with different git versions. I would propose to go with latest and with the minimum git version we want to support. I do not know if we somewhere tell the user a minimum git version. Which version should we add here as minimum version @simonharrer @gregorriegler ?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,11 +5,30 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
+        git-version: [latest, 2.37.0]
     runs-on: ${{ matrix.os }}
     steps:
+    - uses: actions/checkout@v2
+    - name: Install old git version (macOS)
+      if: ${{ matrix.os == 'macos-latest' && matrix.git-version != 'latest'}}
+      run: |
+        brew tap-new mob/local-git
+        brew extract --version=${{ matrix.git-version }} git mob/local-git
+        brew install git@${{ matrix.git-version }} || brew link --overwrite git@${{ matrix.git-version }}
+    - name: Install old git version (ubuntu)
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.git-version != 'latest'}}
+      run: |
+        sudo apt-get update
+        sudo apt-get install gettext asciidoc docbook2x
+        curl https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.37.0.tar.gz --output git-${{ matrix.git-version }}.tar.gz
+        tar -zxf git-${{ matrix.git-version }}.tar.gz
+        cd git-${{ matrix.git-version }}
+        make configure
+        ./configure --prefix=/usr
+        make all info
+        sudo make install install-info
     - name: Show git version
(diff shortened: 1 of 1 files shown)


gregorriegler:
That is such a good idea. I love it. 
Trying to find usage statistics for git versions. Could not find any.
But wikipedia has a nice summary of versions.
https://en.wikipedia.org/wiki/Git
We should probably at least support git starting with 2.17
Everything below that is not maintained anymore by git.

simonharrer:
I would rather reduce the version number until the pipeline breaks to learn what our current min is. Then we can decide what to do. Relying on the maintenance level of git is understandable, but who cares about whether their installed git version is maintained?

gregorriegler:
Ok, so we are now testing what we can. 
On ubuntu, the minimum version that works, 2.37, and the latest.
On mac, I had to omit the minimum version as I could not install it.
I think for now, this is fine. git compile & install takes a lot of time though.
@hollesse please have a final look.

.github/workflows/ci.yaml
```diff
@@ -3,16 +3,39 @@ name: Test
...
     - name: Show git version
       run: git version
-    - uses: actions/checkout@v2
     - name: Use Go 1.16.x
       uses: actions/setup-go@v2
       with:
         go-version: '~1.16.0'
     - name: Test
-      run: go test
+      run: go test -test.v
```
hollesse, line 41: @gregorriegler do we need that -test.v flag? Or was that just for testing which version works and which not?

hollesse:
@gregorriegler I added one comment. Everything else looks fine to me.

.github/workflows/ci.yaml
```diff
@@ -3,16 +3,39 @@ name: Test
 jobs:
   test:
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
gregorriegler: Why would you want it? Are we not failing fast?

Let me tell you why I turned it off. So it was turned on, and what happened was that when one build failed, I didn't get to see the other builds finish. It all suddenly aborted. So I didn't get the feedback where it failed and why. 
It looks like a good way to save hardware resources. On the other hand I don't understand in what way we're not failing fast, if we keep running the other builds after one of them failed. We did see it fail, so we fail as soon as possible, even with the flag set to false, no?.
On python approvals for example, we also use this flag: https://github.com/approvals/ApprovalTests.Python/blob/ee6312d8848e7ee13e8bb5bafbe96bcd7053aa7a/.github/workflows/test.yml#L10

.github/workflows/ci.yaml
gregorriegler: https://github.com/remotemobprogramming/mob/issues/333

.github/workflows/ci.yaml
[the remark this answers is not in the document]
hollesse: I did some research again and github actions seem to be free for all public projects without any limit so this is fine for me.

.github/workflows/ci.yaml
hollesse: fine for me.

PR #335, opened 2022-10-20
Fix cli name detection on Windows

### LINKED ISSUE #322: Mob binary prints as `bin\mob` on Windows
As the title suggests, when the mob binary is printed on a windows system, it reads as `bin/mob` which is incorrect.
Running  `bin/mob` does not work. It should just say `mob`.

Example output:
```
.. To start, including uncommitted changes, use

  bin\mob start --include-uncommitted-changes
```
When you copy the full command and execute it, it won't work.

balintf:
This PR changes `currentCliName` to use `path/filepath` for path manipulation and closes #322 .

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -16,6 +16,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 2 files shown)


hollesse (approved):

hollesse:
Hi @balintf,
thank you for your contribution! The change is well implemented therefore I approve these changes and will merge them :) I am looking forward to your contributions in the future!

balintf:
Thanks for the quick reply and the kind words! :)
I'll try to find a few other issues where I can help out.

hollesse:
@balintf one thing that comes to my mind is the following:
The tests currently don’t work on windows, if mob is not installed, as there is a function called `mobExecutable` which checks if it runs in a Test Environment and then replaces the mob executable with `go run` command. There are two problem with this.
1. the function `isTestEnvironment` does not work correctly on windows as we need to trim the suffix `.exe` from `args[0]`
2. The function `mobExecutable` returns a command which collects all go files in the root folder and passes them to `go run`. This command only works for Linux and Darwin. Not for Windows. There we need to check if we run on windows and return a different command.

do you think that is something you can tackle? Just create an issue with the description above, assign it to yourself and start implementing it ;)

gregorriegler:
https://github.com/remotemobprogramming/mob/issues/336

PR #337, opened 2022-10-25
Fix squash wip tests on Windows

### LINKED ISSUE #336: Tests don't work on Windows because of `mobExecutable()`
There is test-specific production code that provides `mob` as an executable for the test environment.
The command it uses is not compatible with windows:
[mobExecutable()](https://github.com/remotemobprogramming/mob/blob/ee44f3711e70b4a65a224feb316e00d374c0d9fc/squash_wip.go#L71)
```
func mobExecutable() string {
	if isTestEnvironment() {
		wd, _ := os.Getwd()
		return "cd " + wd + " && go run $(ls -1 ./*.go | grep -v _test.go)"
	} else {
		return "mob"
	}
}
```

Q: Why have this command in the first place?
A: mob functions as a `GIT_SEQUENCE_EDITOR` for the squash-wip feature. This is why we have to run mob in order to test squash-wip.
See: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt-sequenceeditor

One Idea would be to add an if clause for windows.

balintf:
This PR fixes #336 by using the `currentCliName` to trim the excess ".exe" part of the executable name in `isTestEnvironment`.
It also fixes the path separators in the command returned by `mobExecutable`, so they work correctly in the git bash environment they are going to get executed at.

### DIFF
--- a/squash_wip.go
+++ b/squash_wip.go
@@ -78,8 +78,9 @@ func mobExecutable() string {
 }
 
 func isTestEnvironment() bool {
-	return strings.HasSuffix(os.Args[0], ".test") ||
-		strings.HasSuffix(os.Args[0], "_test") ||
+	cliName := currentCliName(os.Args[0])
+	return strings.HasSuffix(cliName, ".test") ||
+		strings.HasSuffix(cliName, "_test") ||
 		os.Args[1] == "-test.v"
 }
 
--- a/squash_wip.go
+++ b/squash_wip.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


hollesse:
I am not sure if git bash should be the system we are focusing on for the tests. Why shouldn't we use Powershell? At least for the CI Powershell would make more sense.

gregorriegler:
With git bash, the code doesn't have to change l, and the Windows Runner seems to have git bash installed: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md
If it works, we could give it a try running it in CI.
Why not support it both, git bash and powershell?

balintf:
@hollesse, I think you misunderstood my comment a bit. I am using Powershell to run `go test`, but whenever Git for Windows calls out to an executable (for example the $GIT_EDITOR) it will call it in a git bash shell. This is why we need to change the path separators from `\` to `/`. If we pass single `\` separators git bash will assume they are escape sequences, to fix that we would either need to pass `\\` or `/` in $GIT_EDITOR.

hollesse:
@balintf Oh yes then I got you wrong. Cool then `go test` should run in git bash and powershell. I really like that. Sorry for getting this wrong. I just had a windows computer once in my life and did not use the powershell or git bash at all.

hollesse (approved):

PR #341, opened 2022-10-27
Check if git version is at least 2.13.0

### LINKED ISSUE #332: Add a check for the minimum required git version
As we found out in [#330](https://github.com/remotemobprogramming/mob/pull/330#issuecomment-1277935865) mob can be used with at most git version 2.20.0 and if we ignore the squash-wip feature it can even be used with git 2.13.0. I would like to add a check for the current git version. I am not sure if we should check for 2.13.0 or if we should check for 2.20.0. I would prefere to have a check for mob itself for 2.13.0 and one for the squash-wip feature which should then check for 2.20.0. I would also like to have tests for these version. Therefore we need to check how we could have tests which differs for different git versions. One option to do this could be [go build tags](https://mickey.dev/posts/go-build-tags-testing/)

balintf:
This PR fixes #332 by adding the checks for the required minimum version.
I used a bit more lenient regexp for parsing the git version, since git version strings can be customized during the git commands `./configure` step.

I'm not a 100% sure about adding `minimumGitVersion*` to the const section, adding a single GitVersion struct would have been more elegant, but that can only go in the golbal var section.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -18,6 +18,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -25,7 +26,10 @@ import (
 )
 
 const (
-	versionNumber = "4.0.1"
+	versionNumber          = "4.0.1"
+	minimumGitVersionMajor = 2
+	minimumGitVersionMinor = 13
+	minimumGitVersionPatch = 0
 )
 
 var (
@@ -39,6 +43,48 @@ func openCommandFor(c config.Configuration, filepath string) (string, []string) 
 	}
 	split := strings.Split(injectCommandWithMessage(c.OpenCommand, filepath), " ")
 	return split[0], split[1:]
+}
+
+type GitVersion struct {
+	Major int
+	Minor int
+	Patch int
+}
+
+func parseGitVersion(version string) GitVersion {
+	// The git version string can be customized, so we need a more complex regex, for example: git version 2.38.1.windows.1
+	// "git" and "version" are optional, and the version number can be x, x.y or x.y.z
+	r := regexp.MustCompile(`(?:git)?(?: version )?(?P<major>\d+)(?:\.(?P<minor>\d+)(?:\.(?P<patch>\d+))?)?`)
+	matches := r.FindStringSubmatch(version)
+	var v GitVersion
+	var err error
+	if len(matches) > r.SubexpIndex("major") {
+		v.Major, err = strconv.Atoi(matches[r.SubexpIndex("major")])
+		if err != nil {
+			v.Major = 0
+			return v
+		}
+	}
+	if len(matches) > r.SubexpIndex("minor") {
+		v.Minor, err = strconv.Atoi(matches[r.SubexpIndex("minor")])
+		if err != nil {
+			v.Minor = 0
+			return v
+		}
+	}
+	if len(matches) > r.SubexpIndex("patch") {
+		v.Patch, err = strconv.Atoi(matches[r.SubexpIndex("patch")])
+		if err != nil {
+			v.Patch = 0
+		}
+	}
+	return v
+}
+
+func (v GitVersion) Less(rhs GitVersion) bool {
+	return v.Major < rhs.Major ||
+		(v.Major == rhs.Major && v.Minor < rhs.Minor) ||
+		(v.Major == rhs.Major && v.Minor == rhs.Minor && v.Patch < rhs.Patch)
 }
 
 type Branch struct {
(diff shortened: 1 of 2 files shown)


balintf:
@hollesse, could I ask you to a hacktoberfest-accepted label to this pull request, even if you plan to review and merge it later?
This would be my last PR needed for completing the hacktoberfest challenge.

mob.go
```diff
@@ -188,12 +234,20 @@ func main() {
...
 		say.Error("'git' command was not found in PATH. It may be not installed. " +
 			"To learn how to install 'git' refer to https://git-scm.com/book/en/v2/Getting-Started-Installing-Git.")
 		exit(1)
 	}
 
+	if currentVersion.Less(GitVersion{minimumGitVersionMajor, minimumGitVersionMinor, minimumGitVersionPatch}) {
+		say.Error(fmt.Sprintf("'git' command version '%s' is lower than the required minimum version (%d.%d.%d). "+
+			"Please update your 'git' installation!",
+			versionString, minimumGitVersionMajor, minimumGitVersionMinor, minimumGitVersionPatch))
+		exit(1)
```
simonharrer, line 248: I find this erro very harsh. Could we start with a warning instead?

mob.go
```diff
@@ -1379,12 +1433,14 @@ func gitCommitHash() string {
...
+	_, output, err := runCommandSilent("git", "--version")
 	if err != nil {
-		say.Debug("isGitInstalled encountered an error: " + err.Error())
+		say.Debug("gitVersion encountered an error: " + err.Error())
+		return "", GitVersion{}
 	}
-	return err == nil
+	versionString := strings.TrimSpace(output)
+	return versionString, parseGitVersion(versionString)
 }
```
simonharrer, line 1444: I'd rather have a method `gitVersion` that simply returns the git version string, and if necessary, the caller can then parse this to the `GitVersion` object. Makes it more readible, I think.

mob.go
hollesse: I don’t agree here. If I remember correctly we had issues with basic functionality of mob if you have a older version. The let’s fail fast but then the user definitely sees where the problem is. If we print a warning the user gets another error and maybe find out that his version is to old because there was a warning somewhere at the beginning of the output.

balintf: I have checked what functionality needs 2.13.0, and it is the `git branch --format=...` calls ([git 2.13 release notes](https://github.blog/2017-05-10-git-2-13-has-been-released/#bits-and-bobs). Currently `gitBranches` and `gitRemoteBranches` uses the `--format` parameter.
For example running mob start with this error turned to a warning will result in this:
```
⚠ 'git' command version 'git version 2.12.5' is lower than the required minimum version (2.13.0). Please update your 'git' installation!
  git fetch origin --prune
ERROR git branch --format=%(refname:short)
ERROR error: unknown option `format=%(refname:short)'
ERROR usage: git branch [<options>] [-r | -a] [--merged | --no-merged]
ERROR or: git branch [<options>] [-l] [-f] <branch-name> [<start-point>]
ERROR or: git branch [<options>] [-r] (-d | -D) <branch-name>...
ERROR or: git branch [<options>] (-m | -M) [<old-branch>] <new-branch>
ERROR or: git branch [<options>] [-r | -a] [--points-at]
ERROR
ERROR Generic options
ERROR -v, --verbose         show hash and subject, give twice for upstream branch
ERROR -q, --quiet           suppress informational messages
ERROR -t, --track           set up tracking mode (see git-pull(1))
ERROR --set-upstream        change upstream info
ERROR -u, --set-upstream-to <upstream>
ERROR change the upstream info
ERROR --unset-upstream      Unset the upstream info
ERROR --color[=<when>]      use colored output
ERROR -r, --remotes         act on remote-tracking branches
ERROR --contains <commit>   print only branches that contain the commit
ERROR --abbrev[=<n>]        use <n> digits to display SHA-1s
ERROR
ERROR Specific git-branch actions:
ERROR -a, --all             list both remote-tracking and local branches
ERROR -d, --delete          delete fully merged branch
ERROR -D                    delete branch (even if not merged)
ERROR -m, --move            move/rename a branch and its reflog
ERROR -M                    move/rename a branch, even if target exists
ERROR --list                list branch names
ERROR -l, --create-reflog   create the branch's reflog
ERROR --edit-description    edit the description for the branch
ERROR -f, --force           force creation, move/rename, deletion
ERROR --merged <commit>     print only branches that are merged
ERROR --no-merged <commit>  print only branches that are not merged
ERROR --column[=<style>]    list branches in columns
ERROR --sort <key>          field name to sort on
ERROR --points-at <object>  print only branches of the object
ERROR -i, --ignore-case     sorting and filtering are case insensitive
ERROR exit status 129
```

I think this is not user-friendly error message, while the user will have the warning at the top of the error, more than 90% of the console output is filled with the git output. I would have to agree with @hollesse here about displaying an error and stopping. Although we can reword the error message if you find the wording too harsh.

balintf:
@simonharrer, I've added the requested changes. Only the discussion about the error/warning message remains. If displaying an error is OK we can merge the current version of this PR.

gregorriegler:
I like it. Thanks @balintf

simonharrer:
Yes, thanks for your great work @balintf and patience with my comments!

PR #350, opened 2023-01-06
Fixes Squash wip losing changes without start

### LINKED ISSUE #348: `mob done --squash-wip` before `mob start` loses changes since last `mob start`
(mob.sh v4.0.0)

If you do a `mob done --squash-wip` before doing a `mob start` to get the latest changes, all changes since your last `mob next` are gone and the branch is deleted, making recovery difficult.

Sequence to recreate:

1. Mobber A:
    - `mob start`
    - make change `A1`
    - `mob next`
2. Mobber B: 
    - `mob start`: sees change `A1`
    -  make change `B2`
    - `mob next`
3. Mobber A:
    - `mob done --squash-wip` (forgets to do a `mob start`)
    - branch deleted
    - only change `A1` still exists, change `B2` is lost

Am I missing something? It seems bad for `mob done` to not first do the equivalent of a `mob start` before squashing and deleting a branch. This has happened 3 times now for me—luckily I was able to recover the deleted branch on GitHub. A lot of code could have been lost because, as the facilitator, I didn't participate until many changes had been made (step 2 was two hours of handoffs between other mobbers).

gregorriegler:
Fixes #348 by pulling before squashing.

It's a very simple fix.
What do you think?

Another Idea would be to warn the user that they didn't start or something. 
But that might be more complicated.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -1146,6 +1146,7 @@ func done(configuration config.Configuration) {
 
 	if wipBranch.hasRemoteBranch(configuration) {
 		if configuration.DoneSquash == config.SquashWip {
+			git("pull")
 			squashWip(configuration)
 		}
 		uncommittedChanges := hasUncommittedChanges()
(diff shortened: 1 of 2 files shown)


simonharrer:
pull is problematic. Better option is to do a fetch and a merge with fast forward only - basically what mob start is doing.

Issue #353, opened 2023-01-11
mob start with ci skip option


takaiyuk:
## Context

#329  made `--push-option=ci.skip` available, which allows CI to be skipped in some use cases (perhaps only GitLab).
However, CI is not skipped when `mob start` in my organization, using CircleCI in GitHub.
This is probably because `--push-option=ci.skip` makes no sense except for GitLab.

## Proposal

- Introduce an opt-in option for CI skipping when `mob start`.
  - No change to default behavior due to opt-in option.
- Execute an empty commit with ci-skip message (e.g. `mob start [ci-skip] [ci skip] [skip ci]`) before push for skipping CI.
  - Commits are squashed when `mob done`, so there is no problem with empty commits.

### Implementation Example

- Execution command

```shell
mob start --ci-skip
```

or 

```shell
export MOB_START_WITH_CI_SKIP=true && mob start
```

- Output of command execution

```text
> starting new session from origin/BRANCH-NAME
  git checkout -B mob/BRANCH-NAME origin/BRANCH-NAME
  git commit --allow-empty -m mob start [ci-skip] [ci skip] [skip ci]
  git push --push-option ci.skip --no-verify --set-upstream origin mob/BRANCH-NAME
```

---

## Edit

- Make CI skip option be true as default.
- Initial empty commit with ci-skip message is removed when `mob done --squash-wip`.
  - not removed when `mob done --no-squash`
- `--push-option=ci.skip` to be deleted.

gregorriegler:
Hi @takaiyuk,
Thank you a lot for your contribution!

So if I understand you correctly, the `git push --push-option ci.skip` does not help with skipping the CI in your environment.
However, if you did a `git commit --allow-empty -m mob start [ci-skip] [ci skip] [skip ci]` before, it would?
Is that so? Can you verify that for me?

In that case, I would propose that we skip the configuration property all together, and make the empty commit the default behaviour. As we always would want to skip the CI anyways. There would be no value in turning it off.

takaiyuk:
Hi @gregorriegler, thank you for your quick response.

> So if I understand you correctly, the git push --push-option ci.skip does not help with skipping the CI in your environment.
However, if you did a git commit --allow-empty -m mob start [ci-skip] [ci skip] [skip ci] before, it would?
Is that so? Can you verify that for me?

Yes, you understand correctly.
[...]

PR #354, opened 2023-01-11
mob start with ci skip option

takaiyuk:
## Summary

Implement CI skipping function when `mob start`. 
Check out #353 for details.

## Implementation (editted)

- Skip CI when `mob start`
  - CI is skipped by committing empty with message `mob start [ci-skip] [ci skip] [skip ci]`.
- Remove `--push-option ci.skip`
  - It may by effective only for GitLab at the moment.
  - This requires us to fix test: 5810d58
- Drop the initial empty commit when `mob done --squash-wip`
  - It is removed by `markStartCommitForDropping()`.
  - It is squashed when `mob done` or `mob done --squash`.
  - It is not removed when `mob done --no-squash`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,7 @@ MOB_NOTIFY_COMMAND="/usr/bin/osascript -e 'display notification \"%s\"'"
 MOB_NOTIFY_MESSAGE="mob next"
 MOB_NEXT_STAY=true
 MOB_START_CREATE=false
+MOB_START_WITH_CI_SKIP=true
 MOB_STASH_NAME="mob-stash-name"
 MOB_WIP_BRANCH_QUALIFIER=""
 MOB_WIP_BRANCH_QUALIFIER_SEPARATOR="-"
--- a/README.md
+++ b/README.md
@@ -368,7 +368,6 @@ MOB_NOTIFY_COMMAND="/usr/bin/osascript -e 'display notification \"%s\"'"
 MOB_NOTIFY_MESSAGE="mob next"
 MOB_NEXT_STAY=true
 MOB_START_CREATE=false
-MOB_START_WITH_CI_SKIP=true
 MOB_STASH_NAME="mob-stash-name"
 MOB_WIP_BRANCH_QUALIFIER=""
 MOB_WIP_BRANCH_QUALIFIER_SEPARATOR="-"
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -31,7 +31,6 @@ type Configuration struct {
 	NextStay                       bool   // override with MOB_NEXT_STAY
 	StartIncludeUncommittedChanges bool
 	StartCreate                    bool   // override with MOB_START_CREATE variable
-	StartWithCISkip                bool   // override with MOB_START_WITH_CI_SKIP variable
 	StashName                      string // override with MOB_STASH_NAME
 	WipBranchQualifier             string // override with MOB_WIP_BRANCH_QUALIFIER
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -368,6 +368,7 @@ MOB_NOTIFY_COMMAND="/usr/bin/osascript -e 'display notification \"%s\"'"
 MOB_NOTIFY_MESSAGE="mob next"
 MOB_NEXT_STAY=true
 MOB_START_CREATE=false
+MOB_START_WITH_CI_SKIP=true
```
gregorriegler, line 371: As discussed, I don't think we need this config code.
We want the CI to be skipped everytime.

configuration/configuration.go
```diff
@@ -14,6 +14,8 @@ const (
 	Squash    = "squash"
 	NoSquash  = "no-squash"
 	SquashWip = "squash-wip"
+
+	StartCISkipCommitMessage = "mob start [ci-skip] [ci skip] [skip ci]"
```
gregorriegler, line 18: Feels weird that this constant is in the same place as the other 3 that share a concept. But I don't know how to solve that better now.

configuration/configuration.go
```diff
@@ -62,7 +65,7 @@ func (c Configuration) HasCustomCommitMessage() bool {
 }
 
 func (c Configuration) IsWipCommitMessage(line string) bool {
-	return strings.HasPrefix(line, c.WipCommitMessage)
+	return strings.HasPrefix(line, c.WipCommitMessage) || strings.HasPrefix(line, StartCISkipCommitMessage)
```
gregorriegler, line 68: Do we have a test fail if that `|| strings.HasPrefix(line, StartCISkipCommitMessage)` is ommited?

gregorriegler (commented):
Thank you, well done. 
I think we can omit the configuration code all together, as we always want to skip the CI.
We made the same decision when we implemented the push options.
Then the change becomes much smaller.
Less code = better, less maintenance.

README.md
takaiyuk: OK, I understood but I thought it is convenient for preventing test codes from being broken.
I agree less code is better. I will fix.

configuration/configuration.go
takaiyuk: option 1. Define it as variable.
option 2. Move it to main package because it is no longer used in configure package after deldeting config code.

configuration/configuration.go
takaiyuk: Didn't. `IsWipCommitMessage` is used only in `squashWipGitEditor`, so it does not need to change. I will restore.

configuration/configuration.go
takaiyuk: Or, introduce new type, something like `SquashOption`, to clarify the difference:

```
type SquashOption string

const (
	Squash    = SquashOption("squash")
	NoSquash  = SquashOption("no-squash")
	SquashWip = SquashOption("squash-wip")

	StartCISkipCommitMessage = "mob start [ci-skip] [ci skip] [skip ci]"
)
```

README.md
takaiyuk: Fix: c298999

gregorriegler:
LGTM, can you please increase the version number to `4.2.0` in 
https://github.com/remotemobprogramming/mob/blob/4ac1d3e059ab603bf2464e51934da516e493bb8c/mob.go#L29
and add a changelog entry describing what we did?
I think then we can merge.

takaiyuk:
I did, please fix the changelog message if necessary.

Thanks for discussing and reviewing, @gregorriegler! I am so excited for my PR to be released.

PR #368, opened 2023-03-09
#346 fixed bug that mob thinks there is something to commit if git configuration status.branch is true

### LINKED ISSUE #346: `mob start` breaks when git config `status.branch` is `true`
When the working directory is clean and `status.branch` is `true`, `mob start` returns the error

> &gt; cannot start; clean working tree required

I believe it is caused here, erroneously expecting `git status` to be empty: https://github.com/remotemobprogramming/mob/blob/ee1b922f846c1e9cf31cb0fd501d1a0fa2c012e4/mob.go#L1250-L1253

I believe line 1251 should be calling `git status --porcelain` instead of `git status --short`. I am happy to make a PR for that but do not have the time at the moment to learn enough Go to get an environment running or write a proper test.

hollesse:
closes #346

### DIFF
--- a/mob_test.go
+++ b/mob_test.go
@@ -1749,7 +1749,7 @@ func TestGitVersionCompare(t *testing.T) {
 }
 
 func gitStatus() GitStatus {
-	shortStatus := silentgit("status", "--short")
+	shortStatus := silentgit("status", "--porcelain")
 	statusLines := strings.Split(shortStatus, "\n")
 	var statusMap = make(GitStatus)
 	for _, line := range statusLines {
@@ -2065,6 +2065,8 @@ func createRemoteRepository(path string) {
 	git("--bare", "init")
 	say.Debug("before symbolic-ref")
 	git("symbolic-ref", "HEAD", "refs/heads/"+branch)
+	// see bug #346 changes the output of git status --short
+	git("config", "--local", "status.branch", "true")
 	say.Debug("finished")
 }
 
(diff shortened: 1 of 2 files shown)


mob_test.go
```diff
@@ -1995,6 +1997,8 @@ func cloneRepository(path, remoteDirectory string) {
 	git("clone", "--origin", "origin", "file://"+remoteDirectory, ".")
 	git("config", "--local", "user.name", name)
 	git("config", "--local", "user.email", name+"@example.com")
+	// see bug #346 changes the output of git status --short
+	git("config", "--local", "status.branch", "true")
```
simonharrer, line 2001: Is now every test using this config? Or should we have a single test that makes use of this config change only?

hollesse: I did it for every test and that was good, as I found some more points where `--short` was used and this option let the tests for it fail.

gregorriegler: due to all tests having `git("config", "--local", "status.branch", "true")` now, our test-coverage is not better than it was before that. I would prefer if we had just one test that contained the line `git("config", "--local", "status.branch", "true")` to test specifically that case

hollesse: Do you think this is enough? There where multiple Test that failed. I would at least copy all of them then.

gregorriegler: Did all those tests fail of the same reason? Or did they fail for different reasons?
If they failed for the same reason, I would only pick one

hollesse: They failed because all used git status --short command but they did use it in different methods and for different things

gregorriegler (approved):

benallred:
Would it be helpful for me to attempt a rebase of this branch and resubmit this PR?

PR #369, opened 2023-03-09
#303 better error message if repo does not have any commits

### LINKED ISSUE #303: mob start does not work on git repository without commits
When you clone an empty repository from git and want to start working on it the mob tool throws an error:

image

Currently I do not have a clue what a proper fix for that would be, since the mob tool would need to create a commit which is not developer made to create the corresponding branches.

Maybe it is not worth the effort investigating that.

hollesse:
closes #303

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -250,6 +250,10 @@ func main() {
 	if isGit() {
 		projectRootDir = gitRootDir()
 	}
+	if hasNoCommits() {
+		say.Error("Git repository does not have any commits yet. Please create an initial commit.")
+		exit(1)
+	}
 	configuration := config.ReadConfiguration(projectRootDir)
 	say.Debug("Args '" + strings.Join(os.Args, " ") + "'")
 	currentCliName := currentCliName(os.Args[0])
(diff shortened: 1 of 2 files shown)


gregorriegler (commented):
Is there a reason you chose a command that would yield an error, over a command that would give a successful result?
E.g. `git rev-list --all --count` would count all commits on all branches and yield 0 (no error)

hollesse:
No that was just my knowledge. That's a good idea. Like it :)

gregorriegler:
I made the change @hollesse, please review

mob.go
```diff
@@ -276,10 +278,10 @@ func main() {
 	execute(command, parameters, configuration)
 }
 
-func hasNoCommits() bool {
-	_, _, err := runCommandSilent("git", "rev-list", "-n", "1", "HEAD")
-	return err != nil
-
+func hasCommits() bool {
+	commitCount := silentgit("rev-list", "--all", "--count")
+	println(commitCount)
```
hollesse, line 283: Do we really like to print that? I don‘t think so. And if we would like to print it we should use the say function.

hollesse (commented):
Except for the PrintLn it looks fine to me :)

mob.go
gregorriegler: its a mistake, thanks

gregorriegler: fixed

PR #373, opened 2023-03-29
Directly specify remote ref when pushing.

jstoneham:
Users who have the Git config push.default=tracking, or other values, will have different default behavior on 'git push'.

Example: a newly created mob branch from `origin/mainline` is tracking `origin/mainline` until `--set-upstream` takes effect. The original `git push`, which is intended to push to `origin/mob/mainline`, will actually push the `mob start` commit to `origin/mainline` if `push.default=tracking`.

To avoid this scenario, we will always specify the remote ref we are pushing to.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -1007,7 +1007,7 @@ func startNewMobSession(configuration config.Configuration) {
 	say.Info("starting new session from " + currentBaseBranch.remote(configuration).String())
 	git("checkout", "-B", currentWipBranch.Name, currentBaseBranch.remote(configuration).Name)
 	git("commit", "--allow-empty", "-m", configuration.StartCommitMessage)
-	gitPush(gitHooksOption(configuration), "--set-upstream", configuration.RemoteName, currentWipBranch.Name)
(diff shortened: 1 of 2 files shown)


jakemcc:
:+1: to getting this change landed. We just spent the morning trying to figure out why `mob` kept pushing the newly created branch on top of `origin/master` and it came down to having `push.default = tracking`.

gregorriegler:
Hi @jstoneham, sorry for my late reply.
This is a nice find and I look forward to merge your PR to fix it.
Would you be able to write a test for the behavior you expected but didn't occur?

jstoneham:
I'll try. Sorry to report I'm on vacation at the moment but if I find some time!

gregorriegler:
Oh! Enjoy your vacation 😃. No rush!

jstoneham:
Please take a look at the new test. This test failed before I applied my change. I also did some more testing and removed some of the changes as only the single line is necessary to solve the case I was having trouble with.

gregorriegler:
Thank you very much, @jstoneham !

PR #381, opened 2023-05-05
fix bug where mobs says in a non git directory for mob version you need to be in a git repository

### LINKED ISSUE #380: BUG: Mob v4.4.1 won't display config, help, or version when the current directory is not a git repo
Upgraded to mob v4.4.1 on macOS 12 (via brew).

After updating, the `config`, `help`, `moo`, and `version` commands and switches return an error when the directory is not a git repo. This is a change from v4.4.0.

The following commands return an error message when the directory is not a git repo.

```sh
mob config
mob help
mob moo
mob version
mob --help
mob --version
```

The error message is:

```sh
ERROR expecting the current working directory to be a git repository
```

These commands do not operate on a git repo and should not require a git repo.

(I expect there is a change between 4.4.0 and 4.4.1 that (incorrectly) moved a test for a git repo to be inclusive of all commands.)

hollesse:
closes #380

### DIFF
--- a/mob_test.go
+++ b/mob_test.go
@@ -8,7 +8,6 @@ import (
 	"github.com/remotemobprogramming/mob/v4/test"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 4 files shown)


hollesse:
I refactored the tests and extracted a function run out of the main function to better test the code. So the integration tests where i was really testing the executable are not needed anymore.

mob_test.go
gregorriegler, line 1: What happens if you just run the tests but you don't have built the mob binary beforehand?

hollesse: Just found out my push did not worked. I refactored the tests such that they are not integrationtests anymore

gregorriegler (approved):
nice!

PR #384, opened 2023-05-09
fix: deprecated set-output in GitHub Action workflows

### LINKED ISSUE #379: Deprecation warnings in release build
In our release build we get some deprecation warnings from github. We should fix these, to garantee that our build works in the future.

tobiasehlert:
Due to deprecation of `save-state` and `set-output` commands in GitHub Action workflows, this PR should prevent workflows to fail after the 31st May 2023.

Link to GitHub blogg:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

fix #379

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -50,7 +50,7 @@ jobs:
         if: matrix.os == 'darwin' || matrix.os == 'linux'
         id: shasum
         run: |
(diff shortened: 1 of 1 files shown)


hollesse (approved):
Thank you for your contribution :) this helps us a lot. The change looks good to me :)

PR #386, opened 2023-05-12
fix the logging how to fix uncommited changes

### LINKED ISSUE #383: Include uncommited changes fix does not show previously used parameters
If I have already changes some things and do
`mob start -b green`
then I get an error because I need to include the uncommited changes. The suggested command is `mob start --include-uncommited-changes`. In my opinion the suggestion should be mob start --include-uncommited-changes -b green`.

hollesse:
closes #383

### DIFF
--- a/mob_test.go
+++ b/mob_test.go
@@ -1411,11 +1411,10 @@ func TestStartBranchEnvWithUncommitedChangesFixWithoutBranch(t *testing.T) {
 	output, _ := setup(t)
 
 	setWorkingDir(tempDir + "/local")
-	os.Setenv("MOB_WIP_BRANCH_QUALIFIER", "red")
+	t.Setenv("MOB_WIP_BRANCH_QUALIFIER", "red")
 	createFile(t, "uncommited.txt", "contentIrrelevant")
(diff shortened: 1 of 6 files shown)


mob_test.go
```diff
@@ -1396,6 +1396,54 @@ func TestStartNextPushManualCommits(t *testing.T) {
...
+	runMob(t, tempDir+"/local", "start", "-b", "green")
+
+	assertOutputContains(t, output, "mob start --branch green --include-uncommitted-changes")
+}
+
+func TestStartBranchEnvWithUncommitedChangesFixWithoutBranch(t *testing.T) {
+	output, _ := setup(t)
+
+	setWorkingDir(tempDir + "/local")
+	os.Setenv("MOB_WIP_BRANCH_QUALIFIER", "red")
```
gregorriegler, line 1414: I think we should be using `t.Setenv` instead

gregorriegler (changes requested):
other than my comment it looks good to me

hollesse:
I needed to update the go version to use t.setEnv

gregorriegler (approved):

Issue #387, opened 2023-05-13
Opening last modified file fails when path contains a space


kriber82:
I am using mob.sh on windows inside a project with a space in the project path on disk. When I run `mob start`, instead of opening the last modified file in my current IDEA Window, a new IDEA window opens up.

**Context Info:**

- MOB_OPEN_COMMAND="idea %s" and works fine when removing the space in the project path.
- I am on a windows machine
- The path in my case is "C:/_Development/github/LH-MobSh Kata"

**What happens:**

When I run `mob start` (with last modified file "test.txt"), a new IDEA window opens with two files open:
1.  "LH-MobSh" in "...\_Development\github"
2. "test.txt" in "...\LH-MobSh Kata\Kata"

The log of `mob start --debug` contains the following lines:
```
DEBUG Try to open last modified file
DEBUG Running command <git log -1 --pretty=format:%B> in silent mode, capturing combined output
DEBUG mob next [ci-skip] [ci skip] [skip ci]
DEBUG
DEBUG lastFile:test.txt
DEBUG Running command <git log -1 --pretty=format:%B> in silent mode, capturing combined output
DEBUG mob next [ci-skip] [ci skip] [skip ci]
DEBUG
DEBUG lastFile:test.txt
DEBUG Running command <git rev-parse --show-toplevel> in silent mode, capturing combined output
DEBUG C:/_Development/github/LH-MobSh Kata
DEBUG Starting command idea C:/_Development/github/LH-MobSh Kata/test.txt
DEBUG Open last modified file: C:/_Development/github/LH-MobSh Kata/test.txt
```

**What should happen:**
Preferrably, the space in the path should be handled correctly and the last modified file should be opened in my current IDEA window.
If that's too hard to accomplish, I would hope to see an error message explaining that spaces in the project path (and last modified file names) are not supported. In this case, it would be nice to abort opening the last modified file.

**Preliminary Analysis:**
The main problem seems to be `split := strings.Split(injectCommandWithMessage(c.OpenCommand, filepath), " ")` in `func openCommandFor(c config.Configuration, filepath string) (string, []string)`. This strings.Split operation results in the filepath being split into multiple args.
If you are interested, I could share my attempt on opening the last modified file despite spaces in the project path. It works on my machine and comes with a new test. I will go through the contributing workflow in the next few days, if you are interested.

kriber82:
Added pull request #388

hollesse:
closed with #388

PR #388, opened 2023-05-14
Open last modified file with spaces in path

kriber82:
This is my attempt to allow opening last modified files when the project path on disk contains spaces. I also opened issue  #387 with further details.

The test I added works with a last modified file containing spaces instead of a project path containing spaces, as this was way easier to set up for me and seems to exhibit the same behavior.

As I have no prior experience with Go, I assume my code is not idiomatic. Also, I have not yet extracted functions yet. Therefore, feedback about how my code better fits into the codebase is greatly appreciated.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -41,7 +41,11 @@ func openCommandFor(c config.Configuration, filepath string) (string, []string) 
 	if !c.IsOpenCommandGiven() {
 		return "", []string{}
 	}
-	split := strings.Split(injectCommandWithMessage(c.OpenCommand, filepath), " ")
+	filepathWithoutSpaces := strings.ReplaceAll(filepath, " ", "&spc&")
+	split := strings.Split(injectCommandWithMessage(c.OpenCommand, filepathWithoutSpaces), " ")
+	for i := 0; i < len(split); i++ {
+		split[i] = strings.ReplaceAll(split[i], "&spc&", " ")
+	}
 	return split[0], split[1:]
 }
 
@@ -938,6 +942,9 @@ func openLastModifiedFileIfPresent(configuration config.Configuration) {
 		return
 	}
 	lastModifiedFile := split[1]
+	if strings.Contains(lastModifiedFile, "\"") {
+		lastModifiedFile, _ = strconv.Unquote(lastModifiedFile)
+	}
 	if lastModifiedFile == "" {
 		say.Debug("Could not find last modified file in commit message")
 		return
--- a/mob.go
+++ b/mob.go
@@ -942,7 +942,7 @@ func openLastModifiedFileIfPresent(configuration config.Configuration) {
 		return
 	}
 	lastModifiedFile := split[1]
-	if strings.Contains(lastModifiedFile, "\"") {
+	if strings.HasPrefix(lastModifiedFile, "\"") {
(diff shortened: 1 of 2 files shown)


hollesse:
Hey @kriber82 thank you for your contribution and also the well described issue :) I will have a look at it later, when i have some time. But my first impression is, that it looks good :)

hollesse:
closes #387

mob.go
```diff
@@ -40,7 +40,11 @@ func openCommandFor(c config.Configuration, filepath string) (string, []string)
 	if !c.IsOpenCommandGiven() {
 		return "", []string{}
 	}
-	split := strings.Split(injectCommandWithMessage(c.OpenCommand, filepath), " ")
```
gregorriegler, line 43: I think we should be able to come up with a solution that does not need replacement with an arbitrary placeholder.
E.g. If we split the `OpenCommand` by space first (even before injecting the filepath), and only then inject the filepath into each of the pieces.

mob_test.go
```diff
@@ -875,6 +875,27 @@ func TestStartNextStay_OpenLastModifiedFile(t *testing.T) {
...
+	createFile(t, "file with spaces.txt", "contentIrrelevant")
+	assertOnBranch(t, "mob-session")
+	next(configuration)
+
+	start(configuration)
+
+	assertGitStatus(t, GitStatus{
+		"file with spaces.txt-1": "??",
+	})
+}
```
gregorriegler, line 897: nice test!

gregorriegler (commented):
Thank you @kriber82 for helping improve this tool ❤️.   Looking forward to merge your PR. Please see my review comments where I propose a minor improvement.

kriber82:
I will incorporate your feedback, when I find time, which will probably be sometime next week.

gregorriegler:
Hi @kriber82 do you think you will find the time for making the improvements?

kriber82:
Sorry for the huge delay. Will try to do it within the next days.

mob.go
kriber82: As I see it, splitting `OpenCommand` before injecting the `filepath` would require a special version of `injectCommandWithMessage` that works on the list of pre-split arguments. Otherwise, the "single or no placeholder"-semantics of `injectCommandWithMessage` could not be enforced here.
If we drop this constraint here and assume that there is exactly one %s in OpenCommand, a simple `strings.ReplaceAll` on each split part (command argument) would do the trick, but that would modify existing behavior and duplicate parts of it.

Both of those solutions don't seem better to me than my proposed solution. Do you prefer one of those, or maybe have a better idea how to solve this?

As I contemplated a better solution, I came to the following insights and ideas:
- The problem here seems to be, that `injectCommandWithMessage` is not aware of the fact, that commands may have arguments that contain spaces
- Technically, this could also affect the other calls of `injectCommandWithMessage`
- A cleaner solution would probably have to model the difference between arguments and "words separated by spaces" explicitly.
- This might need to happen as early as parsing the config, as users might want to use double quotes in the config to mark arguments that contain spaces
- I would not be comfortable doing this kind of remodeling in an unfamiliar code-base (mob.sh) and foreign language (go)
- I would, however, be willing to pair- or ensemble-program this with someone more familiar whit both the codebase and go

Which path forward would you prefer?

gregorriegler (approved):

PR #395, opened 2023-07-10
Fix uninstall instructions

dkbnz:
Uninstall instructions are rendering incorrectly on https://mob.sh/

Screenshot 2023-07-10 at 1 41 16 PM

This PR should fix that.


simonharrer:
Thanks for finding and fixing this! :-)

PR #407, opened 2024-01-25
#406 - removed cancelation of mob timers

### LINKED ISSUE #406: Remove cancellation of the local mob timer when doing mob done or mob next
We should remove this feature for the following reasons:

- If you switch from one codebase to another the timer gets stopped and a new timer will start. This could lead to longer rotation times
- the way we stop the timer is just searching for the right process and killing it. This is a behavior which is suspicious for the most virus detection applications and could be one cause why mob is listed for some virus detection tools in windows. (see #357)

hollesse:
This MR closes #406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 4.4.7
+# 4.5.6
 - Removes feature which cancels running timers as this can lead to longer rotations if the codebase is switched. The way it was implemented is also not ideal for virus detection.
 
 # 4.4.6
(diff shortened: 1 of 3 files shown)


gregorriegler (commented):
this is definitely something we should try quickly to get rid of the malware detection.
Since its behaving differently with the timer now, we should at least change the minor version, no?

gregorriegler:
changed version. feel free to merge

hollesse:
I agree it should be a minor version change. But then the Version should be 4.5.0 correct?

gregorriegler:
of course 😃

PR #412, opened 2024-03-15
testpackage should be in test scope

hollesse:
The package `test` and the including file `test.go` weren't in scope test. Therefore this code got bundled into the executable, even if it is not used there. But this code was detected by anti virus sandboxes. Therefore this scope change should be done


gregorriegler:
@hollesse so I tested this again more closely, because I noticed that my IDE has problems finding the functions of the test module if its named like this.

When I change it back to test.go, and I build it and analyze if the test functions actually contained in the binary like the following
```
go clean -cache
go build
go tool nm mob | grep test
``` 
I can not confirm the test functions to be contained. Are you sure they are contained in the binary?

hollesse:
@gregorriegler interesting that your IDE has problems with that. on my machine that works fine. I did not try it the way you did it. What I have done is that I looked at the virus total results of the sandbox mitre tests for the windows executable and there it said that the application is able to write files. I then changed to name of the package as shown here. And then i built a new windows executable and uploaded it again to virus total. Then it did not said that the application is able to write files.

gregorriegler:
I am using IntelliJ Ultimate. It can't find any functions of the test_test.go file.
Marks references as red.
Seems like It thinks test_test.go is a file containing only tests thus doesn't include it.
I'm wondering if this is only a problem of my IDE or has other effects, too.

hollesse:
I had a look again with go tool. Did not know that feature. I think you are right an this changes nothing. So I will just revert it.

PR #413, opened 2024-03-18
#410: added duration to latest commit for a remote branch to mob status

### LINKED ISSUE #410: mob status should show if a mob session exists
It would be helpful when `mob status` shows if a mob session exists on the current branch, when not on a mob branch, to look up if a remote mob branch already exists.

Something like this:

    ➜  myProject git:(main) mob status
    > you are on base branch 'main'
    > wip branch 'mob/main' contains 7 commits. The last 5 were:
    > 38209edb 6 hours ago <Wanda Hess>
    > 2359e605 5 hours ago <Ross Carter>
    > 3868b5c6 5 hours ago <Claudia Osborne>
    > b848355a 5 hours ago <Lee Klein>
    > 209e8bb2 5 hours ago <Ammar Carlson>

hollesse:
closes #410 

the result of the change can be seen here. 
```
$ mob status
> you are on base branch 'main'
> remote wip branches detected:
  - origin/mob/main (10 months ago)
  - origin/mob/main-my-feature (1 year, 4 months ago)

```

if there is no remote wip branch:
```
$ mob status
> you are on base branch 'main'
> no remote wip branches detected!
```

### DIFF
--- a/status_test.go
+++ b/status_test.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	config "github.com/remotemobprogramming/mob/v4/configuration"
+	"strconv"
+	"testing"
+)
+
+func TestExecuteKicksOffStatus(t *testing.T) {
+	output, _ := setup(t)
+
+	execute("status", []string{}, config.GetDefaultConfiguration())
+
+	assertOutputContains(t, output, "you are on base branch 'master'")
+}
+
+func TestStatusMobProgramming(t *testing.T) {
+	output, configuration := setup(t)
+	start(configuration)
+
+	status(configuration)
+
+	assertOutputContains(t, output, "you are on wip branch mob-session")
+}
+
+func TestStatusWithMoreThan5LinesOfLog(t *testing.T) {
+	output, configuration := setup(t)
+	configuration.NextStay = true
+	start(configuration)
+
+	for i := 0; i < 6; i++ {
+		createFile(t, "test"+strconv.Itoa(i)+".txt", "contentIrrelevant")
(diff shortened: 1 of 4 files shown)


status_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	status(configuration)
+	// 6 wip commits + 1 start commit
+	assertOutputContains(t, output, "wip branch 'mob-session' contains 7 commits.")
+}
+
+func TestStatusDetectsWipBranches(t *testing.T) {
+	output, configuration := setup(t)
+	start(configuration)
```
hollesse, line 43: I would like to adjust the test that it does not depend on the empty commit when doing mob start. As I would like to remove this empty commit again.

gregorriegler: ok, great!

status_test.go
```diff
@@ -41,7 +42,10 @@ func TestStatusWithMoreThan5LinesOfLog(t *testing.T) {
 func TestStatusDetectsWipBranches(t *testing.T) {
 	output, configuration := setup(t)
 	start(configuration)
+	createFile(t, "test.txt", "contentIrrelevant")
+	next(configuration)
 	git("checkout", "master")
+	time.Sleep(2 * time.Second)
```
gregorriegler, line 48: ouch, do we really need that?

hollesse: I think so. On my machine i ran into the problem that it was exactly `one second ago` and the test failed. One other option would be to just assert `ago` or `second` what do you think?

PR #414, opened 2024-03-24
#394 added cli option --room

### LINKED ISSUE #394: Allow specify mob timer room by cli option
We usually mob in the same room and set up our mob config using `MOB_TIMER_ROOM`.

However sometimes we split and fall back to pair programming or similar. 
In this cases it would be nice if we could overwrite the mob timer room using the cli when we start a mob session using `mob start`.

hollesse:
closes #394

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -144,6 +144,12 @@ func ParseArgs(args []string, configuration Configuration) (command string, para
 			newConfiguration.StartCreate = true
 		case "--delete-remote-wip-branch":
 			newConfiguration.ResetDeleteRemoteWipBranch = true
+		case "--room":
+			if i+1 != len(args) {
+				newConfiguration.TimerRoom = args[i+1]
+			}
+			i++ // skip consumed parameter
+
 		default:
 			if i == 1 {
 				command = arg
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -72,6 +72,39 @@ func TestParseArgsMessage(t *testing.T) {
 	test.Equals(t, "ci-skip", configuration.WipCommitMessage)
 }
 
+func TestParseArgsStartRoom(t *testing.T) {
+	configuration := GetDefaultConfiguration()
+	test.Equals(t, configuration.WipBranchQualifier, "")
+
+	command, parameters, configuration := ParseArgs([]string{"mob", "start", "--room", "testroom"}, configuration)
+
+	test.Equals(t, "start", command)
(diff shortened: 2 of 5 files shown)


configuration/configuration.go
```diff
@@ -144,6 +144,12 @@ func ParseArgs(args []string, configuration Configuration) (command string, para
 			newConfiguration.StartCreate = true
 		case "--delete-remote-wip-branch":
 			newConfiguration.ResetDeleteRemoteWipBranch = true
+		case "--room":
+			if i+1 != len(args) {
+				newConfiguration.TimerRoom = args[i+1]
+			}
```
gregorriegler, line 150: This block now has 3 duplications --branch as well as --message

configuration/configuration_test.go
```diff
@@ -72,6 +72,39 @@ func TestParseArgsMessage(t *testing.T) {
...
+	configuration := GetDefaultConfiguration()
+	test.Equals(t, configuration.WipBranchQualifier, "")
+
+	command, parameters, configuration := ParseArgs([]string{"mob", "timer", "open", "--room", "testroom"}, configuration)
+
+	test.Equals(t, "timer", command)
+	test.Equals(t, "open", strings.Join(parameters, ""))
+	test.Equals(t, "testroom", configuration.TimerRoom)
+}
+
```
gregorriegler, line 107: nice, lets also test that the arg is doing what its supposed to be doing

configuration/configuration.go
hollesse: The following function could solve this issue:
```go
func SetNextArg(args []string, i *int, setValue func(value string)) {
	if *i+1 < len(args) {
		setValue(args[*i+1])
		*i += 1
	}
}
```

We could use it in the code as follows:
```go
SetNextArg(args, &i, func(value string) {
	newConfiguration.TimerRoom = value
})
```

This solution is in my opinion much more complex and therefore i would stay with 3 code duplications here.

configuration/configuration_test.go
hollesse: I would not extend the test here to check if the timer works. As this is just a test which should check if the configuration works. We have a test in `mob_test.go` which checks it the browser would open up, see `mob_test.go:1726`:
```go
func TestOpenTimerInBrowserWithTimerRoom(t *testing.T) {
	mockOpenInBrowser()
	output, configuration := setup(t)
	configuration.TimerRoom = "testroom"

	err := openTimerInBrowser(configuration)

	assertOutputNotContains(t, output, "Timer Room is not configured.")
	assertNoError(t, err)
}
```

PR #415, opened 2024-03-24
344 fixed timer bug

### LINKED ISSUE #344: Mob start timer starts timer with zero minutes if you add a parameter which is not a number
When I do `mob timer notanumber` this will start a timer with 0 Minutes and directly say `mob next`. This behavior also applies to `mob start notanumber`. 

In my opinion it should work like this:

## `mob timer notanumber` 
- should fail with an error message

## `mob start notanumber``
- If the configuration `MOB_TIMER` is set, this value should be used instead and a warning should be printed
- if configuration `MOB_TIMER` is not set, a warning should be printed

## `mob start` and `MOB_TIMER="notanumber"`
- a warning should be printed

@simonharrer please add your feedback!

hollesse:
closes #344

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -2,19 +2,11 @@ package main
 
 import (
 	"bufio"
-	"bytes"
-	"crypto/tls"
-	x509 "crypto/x509"
-	"encoding/json"
 	"errors"
 	"fmt"
 	config "github.com/remotemobprogramming/mob/v4/configuration"
 	"github.com/remotemobprogramming/mob/v4/help"
-	"github.com/remotemobprogramming/mob/v4/open"
 	"github.com/remotemobprogramming/mob/v4/say"
-	"io"
-	"net/http"
-	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


mob.go
```diff
@@ -310,10 +302,13 @@ func execute(command string, parameter []string, configuration config.Configurat
 			return
 		}
 		if len(parameter) > 0 {
-			timer := parameter[0]
-			startTimer(timer, configuration)
+			if err := startTimer(parameter[0], configuration); err != nil {
```
gregorriegler, line 305: I see duplication

mob.go
```diff
@@ -310,10 +302,13 @@ func execute(command string, parameter []string, configuration config.Configurat
...
+				exit(1)
+			}
```
gregorriegler, line 307: This block has 5 duplications

mob.go
```diff
@@ -507,189 +491,6 @@ func executeCommandsInBackgroundProcess(commands ...string) (err error) {
...
-	body, responseReadingErr := io.ReadAll(response.Body)
-	if responseReadingErr != nil {
-		return fmt.Errorf("failed to read the http response: %w", responseReadingErr)
-	}
-	if string(body) != "" {
-		say.Info(string(body))
-	}
-	return nil
-}
-
```
gregorriegler, line 692: this is really hard to review becaues it is just one commit that changes everything. I dont know whether it was just moved, or wether something changed there, too

gregorriegler:
feel free to merge

mob.go
hollesse: Oh yes got your point. Should have done this in two seperate commits. I think i will just do it again and split it in two commits.

hollesse:
Please see #416

PR #421, opened 2024-04-16
Exit

### LINKED ISSUE #402: Exits with 0 on at least some failures
Here's one: `cannot start; clean working tree required`

But exits with 0.

hollesse:
Closes #402. I needed to add the mock exit functionality here as the exiting leads to the test failing. If you have a better idea please let me know.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -300,7 +300,7 @@ func execute(command string, parameter []string, configuration config.Configurat
 	case "s", "start":
 		err := start(configuration)
 		if !isMobProgramming(configuration) || err != nil {
-			return
+			exit(1)
 		}
 		if len(parameter) > 0 {
 			timer := parameter[0]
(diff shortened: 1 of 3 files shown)


gregorriegler (commented):
its ok just trying to think of a way that helps us avoid contaminating our tests with such setup and teardown functionality. Did you try TestMain?
https://blog.marcnuri.com/go-setting-up-and-tearing-down-unit-tests

hollesse:
> its ok just trying to think of a way that helps us avoid contaminating our tests with such setup and teardown functionality. Did you try TestMain? https://blog.marcnuri.com/go-setting-up-and-tearing-down-unit-tests

Test Main would not be the right thing here. I want to explicitly set this, as I am expecting here that the application terminates. If I would use test main I would need to check everytime if there is the exit message in the output to check if the tools terminates where it should not terminate. We could think about having a testrunner for that, but I am not sure if this really introduces this tests, as it hides this behavior. 

So what do you think about that @gregorriegler

mightyiam:
Thank you

PR #424, opened 2024-04-27
Refactor sayFixUncommittedChanges

gregorriegler:
Improving this so I can wrap my head around it when creating a thefuck rule


hollesse (approved):
Really like the way you have done it :)

PR #425, opened 2024-04-29
Add reference to DB Systel in "Who is using 'mob'?"

Morl99:
Let's add our usage to the Readme (as discussed in a conversation with @simonharrer during the JAX in Mainz).


hollesse:
@Morl99 I love to see my old employer in the list of users 😉 Thank you for your contribution ❤️ If you like join our Slack community. You can join [here](http://slack.mob.sh)!

PR #427, opened 2024-05-15
Fix build on non-Linux Unix (such as NetBSD).

schmonz:
Build-tested on macOS 14.5, Ubuntu 24.04, NetBSD 10.0.


gregorriegler:
Thank you very much @schmonz !

Issue #372, opened 2023-03-27
Mob command visual "feedback"


mpgoodwin-wsa:
Would it be possible that the mob tool could print "git fetch" before doing the fetch, rather than printing these status messages after the command has completed. When working on a slow connection, 'mob start' looks like it is just hung.

PR #428, opened 2024-06-24
Print Git Commands before they are run

gregorriegler:
Addresses https://github.com/remotemobprogramming/mob/issues/372

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -1138,6 +1138,7 @@ func gitWithoutEmptyStrings(args ...string) {
 }
 
 func git(args ...string) {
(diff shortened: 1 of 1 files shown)


simonharrer:
Be aware that the git command might be printed two times in case of error

hollesse (approved):
Like that. If we get the log twice in case of error does not hurt!

gregorriegler:
I agree, it does not hurt.
If we want it only a single time, we need to either change the model to do something like
1. git fetch
2. git fetch ... ✅
or we chose to update the terminal with something like uilive, but I would rather not use a third party library.

PR #435, opened 2024-07-05
fix #434 ci skip push option is evaluated incorrectly

### LINKED ISSUE #434: MOB_SKIP_CI_PUSH_OPTION_ENABLED is evaluated incorrectly
The new config option MOB_SKIP_CI_PUSH_OPTION_ENABLED is evaluated incorrectly.
When set to true, no push option is appended.
When set to false, the push option "ci.skip" is appended.
It should be the other way around, right?

see mob.go func gitPushArgs:

```
func gitPushArgs(c config.Configuration) []string {
	pushArgs := []string{"push"}
	if c.SkipCiPushOptionEnabled {
		return pushArgs
	} else {
		return append(pushArgs, "--push-option", "ci.skip")
	}
}
```

P.S.: The new option is not yet documented in the README.

hollesse:


stefanscheidt (approved):
Looks good to me. Thanks for the quick fix.

gregorriegler (commented):
lgtm

PR #440, opened 2024-08-24
Fix missing directory

michelgrootjans:
TestStartDespiteGitHook failed on my laptop.

Cause: `/.git/hooks` does not get created by default.

Fix: during tests, create directories before adding files to it.


michelgrootjans:
There seems something wrong with the virus scanner service. The log of the github action mentions:
```
Run crazy-max/ghaction-virustotal@v3
4 file(s) will be sent to VirusTotal for analysis
Error: Cannot send mob-virustotal-windows.exe to VirusTotal: AxiosError: Request failed with status code 502
```

see: https://github.com/remotemobprogramming/mob/actions/runs/10540106680/job/29204408155?pr=440

PR #447, opened 2024-09-20
#446 mob start --discard-uncommitted-changes

### LINKED ISSUE #446: mob start --discard-uncommitted-changes
Hi. How about a new start command option [--discard-uncommitted-changes|-d]   Discard uncommitted changes before switching to wip branch ?

Motivation: While not being the typist, I quite often try things out in the code and then forget to do a git reset --hard. Thus I have to type
```
> mob s 10 // failing due to uncommitted changes
> git reset --hard
> mob s 10
```
With this feature, I can simply do
```
> mob s 10 // failing due to uncommitted changes
> mob s 10 -d
```
or - knowing myself ...
`> mob s 10 -d`

Original author: @stefanscheidt 
Source: https://mobsh.slack.com/archives/C06UM83BKPX/p1726815243824169

hollesse:
This PR closes #446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # 5.2.0
 - Feature: `mob done` now pulls when someone else already did `done`
 
+# 5.2.0
+- Feature: `mob done` now pulls when someone else already did `done`
+
 # 5.1.1
 - Fix: `mob break 5` is now correctly parsed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 5.3.0
+- Feature: `mob start --discard-uncommitted-changes` allows to discard uncommitted changes and then starting a new mob session
+
+Thank you @stefanscheidt for this feature request
+
 # 5.2.0
 - Feature: `mob done` now pulls when someone else already did `done`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,6 @@ Thank you @stefanscheidt for this feature request
 # 5.2.0
 - Feature: `mob done` now pulls when someone else already did `done`
 
-# 5.2.0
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+# 5.3.0
+- Feature: `mob start --discard-uncommitted-changes` allows to discard uncommitted changes and then starting a new mob session
+
+Thank you @stefanscheidt for this feature request
+
+# 5.2.0
```
gregorriegler, line 6: sorry, we released 5.2.0 on socratesAT, we need to remove this paragraph :)

mob.go
```diff
@@ -527,7 +527,7 @@ func deleteRemoteWipBranch(configuration config.Configuration) {
 
 func start(configuration config.Configuration) error {
 	uncommittedChanges := hasUncommittedChanges()
-	if uncommittedChanges && !configuration.StartIncludeUncommittedChanges {
+	if uncommittedChanges && !configuration.StartIncludeUncommittedChanges && !configuration.StartDiscardUncommittedChanges {
```
gregorriegler, line 530: as you noticed by implementing this feature, its a flag for the user - but for us internally, its not a flag, but one of many options how to deal with uncommitted changes. So I think we should also code it this way. 

the flags should result in some sort of enum 
```
DealWithUncommittedChanges {
    Include,
    Discard,
    Ignore
}
```
where each of it leads to another path in our code.

This also eliminates the invariant where both booleans could be true (which doesnt make any sense).

The names I used in the enum are just quickly made up. You might find better names, or a better solution to implement the polymorphism, but please don't use two booleans for the same decision point.

hollesse: @gregorriegler I think you are right. It should be a enum. I adjusted it. Please have a look again :)

gregorriegler: i like it! feel free to merge

PR #453, opened 2024-10-06
Fix #452 installed binary path check on Windows git bash

### LINKED ISSUE #452: Test if installation is successful fails
The "How to install" states:

```
# works for macOS, linux, and even on windows in git bash
curl -sL install.mob.sh | sh
```

The installation on windows in git bash seems to be successful, but the test reports the contrary:

```
$ curl -sL install.mob.sh | sh
installing latest 'mob' release from GitHub to C:/Users/johndoe/bin...
Mob binary location: /c/Users/johndoe/bin/mob
Mob binary version: v5.3.1
(!) The installation location doesn't match the location of the mob binary.
    This means that the binary that's used is not the binary that has just been installed
    You probably want to delete the binary at /c/Users/johndoe/bin/mob
```

It states the installation was not successful (but should report success as it seems):
```
    This means that the binary that's used is not the binary that has just been installed
```

```
$ mob --version
v5.3.1
```

elliotwutingfeng:
Closes #452

When installing on Windows via git bash, `command -v mob` gives a UNIX style path while `$target` gives a UNIX style path but with a Windows style drive lettering (e.g. "C:/..."), leading to a mismatch.

This patch checks if the system is Windows, and if so, converts the expected path to the same format (Windows style drive lettering, forward slashes) before comparison.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -175,7 +175,12 @@ check_say() {
 
 check_installation_path() {
   location="$(command -v mob)"
-  if [ "$location" != "$target/mob" ] && [ "$location" != "" ]; then
+  target_location="$target/mob"
+  if [ "$(determine_os)" == "windows" ]; then
+    location="$(cygpath -m "$location")"
+    target_location="${target_location}.exe"
+  fi
+  if [ "$location" != "$target_location" ] && [ "$location" != "" ]; then
     echo "(!) The installation location doesn't match the location of the mob binary."
     echo "    This means that the binary that's used is not the binary that has just been installed"
     echo "    You probably want to delete the binary at $location"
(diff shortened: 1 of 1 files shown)


gregorriegler (changes requested):
Thank you very much for looking into this issue.
Unfortunately the proposed solution only works if cygwin is installed. We need this to work regardless.

See https://chatgpt.com/share/6702329e-a264-800f-a7cb-706fccee0716

elliotwutingfeng:
Current iteration requires `sed`. MINGW64 does not include any string manipulation tools by default. However, `install.sh` assumes that `curl`, `grep`, `cut` and `tr` are available, so there is a good chance `sed` would be available as well.

gregorriegler (approved):

PR #457, opened 2024-10-10
Inject http.client into httpclient.SendRequest

Alex-CD:
This is a tiny change that refactors httpclient to:

1. Remove `getHttpClient` (which was unused).
2. Inject a `*http.client` into `SendRequest`, rather than having `SendRequest` self-initialise one.
3. Have usages of `SendRequest` initialise a `http.client` using `GetHttpClient`

In terms of logic, things are the same as they were before, with the exception of the request being initialized _after_ the client, instead of before.

Now that `SendRequest` is decoupled from `http.DefaultClient`, it's a little easier to write unit tests for it (which I'd be happy to do)

I'd be super interested in your thoughts on the general theme of doing dependency injection like this.

### DIFF
--- a/goal/goal.go
+++ b/goal/goal.go
@@ -64,10 +64,8 @@ func putGoalHttp(goal string, configuration config.Configuration) error {
 	if err != nil {
 		return err
 	}
-
 	httpClient := httpclient.GetHttpClient(configuration.TimerInsecure)
 	_, err = httpclient.SendRequest(requestBody, "PUT", getGoalUrl(configuration), httpClient)
-
 	return err
 }
 
--- a/goal/goal.go
+++ b/goal/goal.go
@@ -64,8 +64,8 @@ func putGoalHttp(goal string, configuration config.Configuration) error {
 	if err != nil {
 		return err
 	}
-	httpClient := httpclient.GetHttpClient(configuration.TimerInsecure)
-	_, err = httpclient.SendRequest(requestBody, "PUT", getGoalUrl(configuration), httpClient)
+	client := httpclient.CreateHttpClient(configuration.TimerInsecure)
+	_, err = client.SendRequest(requestBody, "PUT", getGoalUrl(configuration))
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


gregorriegler (changes requested):
Hi @Alex-CD .
Thank you for your effort.
Please bear with me, but I would like the following approach.

I suggest that in order to distinguish between net/http.Client and our http.Client we need better naming.
I suggest to prefix the official http.Client with net: netHttpClient

In order to test our code we need our own wrapper of that netHttpClient, that implements an interface, that also has a Spy implementation.
Such as this code would do: 

```
type Client interface {
	SendRequest(method string, url string, body []byte) error
}

type HttpClient struct {
	netHttpClient *http.Client
}

func CreateHttpClient(disableSSLVerification bool) HttpClient {
	return HttpClient{
		netHttpClient: GetNetHttpClient(disableSSLVerification),
	}
}

func GetNetHttpClient(disableSSLVerification bool) *http.Client {
	if disableSSLVerification {
		transCfg := &http.Transport{
			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
		}
		return &http.Client{Transport: transCfg}
	}
	return http.DefaultClient
}

type HttpClientSpy struct {
	// used in unittests
	// records calls to http, but does not send them
}

func CreateHttpClientSpy() HttpClientSpy {
	return HttpClientSpy{
		// ...
	}
}

func (c HttpClient) SendRequest(requestMethod string, requestUrl string, requestBody []byte) error {
	say.Info(requestMethod + " " + requestUrl + " " + string(requestBody))
	// ...
}
```

We never ever use the returned responseBody, so I removed that from the return value.
When we run mob in a unittest, we would just use the Spy implementation.
The implementation of our own wrapper (this can be a spy, or the real one) needs to be either passed down the call stack, or stored somewhere.
I dont yet know whats the best option here. Would start with passing it down the call stack.

A Unittest would look something like this (pseudocode):

```
func TestBreakOf5(t *testing.T) {
	_, configuration := setup(t)
	httpClient = createHttpClientSpy()

	break(httpClient)

	httpClientSpy.assertCalledWith("GET", "url", "{the expected json}")
``` 

Since these are quite some changes, i propose we do it in small steps.

You would have to undo the parameter injection of netHttpClient.
We don't want any code other than our httpclient package, to depend on netHttpClient.
After you undid the injection, i would already merge.

If you like, you could also start preparing the interface and HttpClient Object.

What are your thoughts on this approach?

Alex-CD:
I'm fully on board with this approach :) 
I've pushed an update that removes the dependency injection and gets the initial HttpClient and Client interface.

Alex-CD:
I also see what you mean about the naming being confusing -- to that end I've changed some variable names in what I've added ([example](https://github.com/remotemobprogramming/mob/pull/457/files#diff-4e6594f9ef6920412f75106fea39ffbbba55afc6bc04c0fda182aa7940ea0f99R160))

gregorriegler (approved):

PR #459, opened 2024-10-11
fix #451

### LINKED ISSUE #451: "mob start" fails when base branch isn't a local branch
I'm not sure if this is a feature or a bug: Let's assume an ensemble works on a task with the base branch some-task. Now a colleague joins, clones the repo, switches immediately to mob/some-task (!) and enters mob s 10. Then he/she will see:

```
  git fetch origin --prune
ERROR git rev-list --count --left-only refs/heads/some-task...refs/remotes/origin/some-task
ERROR fatal: ambiguous argument 'refs/heads/some-task...refs/remotes/origin/some-task': unknown revision or path not in the working tree.
ERROR Use '--' to separate paths from revisions, like this:
ERROR 'git <command> [<revision>...] -- [<file>...]'
ERROR exit status 128
```

Of course he/she should have switched to some-task and do a mob s 10 (or probably mob s 10 --join ).
Is the current behaviour intended or could this be improved?

hollesse:


mob.go
```diff
@@ -551,6 +566,11 @@ func start(configuration config.Configuration) error {
 
 	createRemoteBranch(configuration, currentBaseBranch)
 
+	if !currentBaseBranch.hasLocalBranch(localBranches) {
+		silentgit("checkout", "-b", currentBaseBranch.Name)
+		silentgit("checkout", currentBranch.Name)
```
simonharrer, line 571: why do we need this line? It feels wrong

hollesse: Hi @simonharrer, first I thought we need it, because the git command which is used to check if there are unpushed commits. But because of your comment I thought more about it and if there is no local base branch there can not be unpushed commits. but after implementing this change the test failed again. In the function `sayLastCommitsList` we look at the local basebranch. I now implemented it the way that we try this behavior if it fails we try to check against remote. Let me know if you think it is better to directly compare to the remote base branch instead of trying to compare to the local branch first. I think it could make sense but was not completely sure about it.

mob.go
```diff
@@ -1001,12 +1016,16 @@ func squashOrCommit(configuration config.Configuration) string {
 	}
 }
 
-func sayLastCommitsList(currentBaseBranch string, currentWipBranch string) {
-	commitsBaseWipBranch := currentBaseBranch + ".." + currentWipBranch
```
hollesse, line 1005: @simonharrer this is the place where we could also just use the currentRemoteBaseBranch, this would be a much easier solution.

mob.go
```diff
@@ -122,6 +122,19 @@ func (branch Branch) hasRemoteBranch(configuration config.Configuration) bool {
 	return false
 }
 
+func (branch Branch) hasLocalBranch(localBranches []string) bool {
```
gregorriegler, line 125: Imo, hasLocalBranch should not need an argument localBranches. It can fetch them itself. And if local branches are fetched multiple times, I still think its worth it. When I need to find out how to check if a localBranch exists in our code, and I find this function, I don't want it to make me search for yet another function.

gregorriegler: The same is probably applicable to determineBranches

PR #460, opened 2024-10-14
Check string equality more portably.

schmonz:
`==` for string equality is non-POSIX. One `=` suffices.


hollesse:
Hi @schmonz,
thank you for this improvement!

PR #462, opened 2024-12-10
Replace `--delete-remote-wip-branch` with a prompt for the user on whether they would like to continue

WiaanB:
Replace `--delete-remote-wip-branch` with a prompt for the user on whether they would like to continue

Reason: To make the flow a little more user friendly, by repeatedly running into the issue of having to rerun the command with the flag

- Kept User Config in place to retain default behaviour
- Updated the tests

### DIFF
--- a/ask/ask.go
+++ b/ask/ask.go
@@ -1,0 +1,32 @@
+package ask
+
+import (
+	"bufio"
+	"github.com/remotemobprogramming/mob/v5/say"
+	"io"
+	"os"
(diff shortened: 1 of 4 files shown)


simonharrer:
Violates https://github.com/remotemobprogramming/mob?tab=readme-ov-file#design-concepts

> mob is not interactive.

If we would go towards being interactive, this would be a massive change in the design of the tool. @hollesse what do you think?

PR #466, opened 2025-03-21
feat: add shortcut for `--create` as `-c`

TimsDevCorner:
I've added a shortcut for `mob start --create` as `mob start -c`

The "configuration/configuration_test.go" has not been executed with `go test -v`, thus I updated the README with `go test ./...` to run tests in all modules.

Regarding the version update: I was not sure, how your updates work exactly. I've included it in this PR, feel free to remove or update, as you see fit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 5.3.4
+- Feature: Add shortcut for `mob start --create` as `mob start -c`.
+
 # 5.3.3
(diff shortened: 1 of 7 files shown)


TimsDevCorner:
What shall I do about the VirusTotal Scan?

hollesse:
Hi @TimsDevCorner First of all thank you for your contribution. I have seen this MR and probably will merge it today. For know you can ignore the virus total check. It seems that there were some API changes and we therefore need a new version of this check. I will fix this before merging your MR.

hollesse (approved):

PR #467, opened 2025-03-31
Update Homebrew installation and fix a few typos

SimplyDanny:
Now that `mob` is available from the official Homebrew repository, we can recommend to install it from there. Works for Linux and macOS.


gregorriegler:
Thank you!

PR #469, opened 2025-06-27
Avoid git fetch'ing all branches by default

hjwp:
On a repo with a significant number of remote branches, fetching all remotes takes a substantial amount of time (literal minutes in our case where there are thousands of remotes)

This changes each `git fetch` call to fetch only the current branch name, and has specialcase handling for the `mob start --create` option where we know we want to create a brand new branch.

Appreciate this might not be a problem for most people!

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -550,8 +550,9 @@ func start(configuration config.Configuration) error {
 	}
 
 	currentBranch := gitCurrentBranch()
-	git("fetch", configuration.RemoteName, currentBranch.Name)
-
+	if !configuration.StartCreate {
+		git("fetch", configuration.RemoteName, currentBranch.Name)
+	}
 	currentBaseBranch, currentWipBranch := determineBranches(currentBranch, gitBranches(), configuration)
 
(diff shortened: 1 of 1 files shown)


hollesse:
Hi thank you for your contribution. The tests do not successfully complete. But I even think that just fetching the current branch is not a solution. At least you would have to check for the correlated mob branch if you are on a base branch and vice versa. And even then I am not sure if that would work right now. 

Could you explain to me why the repo of your Project has so many things that get fetched? I want to understand the background behind it.

hjwp:
thanks for taking an interest!  the repo i'm working on currently has over 4000 open branches, and at my last count had 178 merges in the last 4 hours, ie 178 branches closed/deleted.  that makes the overhead of an unqualified `git pull` or `git fetch` prohibitive, hence wanting to only pull/fetch the current branch, if i can.

hjwp:
sorry for not finding the tests!  I've never written any Go before so i'm finding my feet.

> just fetching the current branch is not a solution. At least you would have to check for the correlated mob branch if you are on a base branch and vice versa. And even then I am not sure if that would work right now.

can you tell me a bit more about why it wouldn't work?

PR #482, opened 2026-02-20
feat: Add ci.skip push option to mob next

heidiberry:
Users can configure MOB_SKIP_CI_PUSH_OPTION_ENABLED so the ci.skip push option is added when running mob start.

This pull request extends this configuration to include ci.skip when running mob next.

### DIFF
--- a/mob.go
+++ b/mob.go
@@ -31,7 +31,7 @@ const (
 
 var (
 	gitClient = &mobgit.Client{}
-	args []string
+	args      []string
 )
 
 func openCommandFor(c config.Configuration, filepath string) (string, []string) {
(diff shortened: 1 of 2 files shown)


philwolstenholme:
Thanks for making this MR @heidiberry :) 

Heidi and I work at the same org but in different teams. My team uses mob.sh daily, and Heidi's team manages the CI infrastructure that is currently working overtime because our `mob next` pushes aren't using the `ci.skip` option. 

@hollesse other than a green CI run once the workflow has been approved, is there anything else you'd like to see from this before it could be merged? I don't mind contributing any docs changes or extra tests if you'd like to see them.

hollesse:
Hi @philwolstenholme and @heidiberry 

Thank you for your contribution. We added the push option, when creating the new mob branch, as we did not have a proper way to skip ci with an adjusted commit message here. In the case of mob next, we have a commit message, which includes different kind of ci skip. Here you can see how the default mob next commit message looks like: "mob next [ci-skip] [ci skip] [skip ci]" Does your CI system does not support skiping CI job with one of these text fragments in the commit message? 

Right now I did not have enough time to think about you suggested changes and the potential impact. Therefore please have a look if you could instead just use the ci-skip within the commit message, as a workaround.

